PR #84, opened 2018-11-15
Rename everything (hopefully) to Charcoal

osanoj:
# Why?
New better name needs to be reflected "everywhere".

# What?
Everything I could think of and find is now renamed, but there might be more things.


3lvis:
Yolo approval!

PR #89, opened 2018-11-21
Submarkets data

osanoj:
# Why?
Filter setup needs to handle submarkets.

# What?
Expand enum configuration with nested enums. This might not be the best for when all markets should be fully configured, but it gives us something to work with.

# Show me
Not much to see.


osanoj:
Soon there will be another PR about switching submarkets that add functionality to this

osanoj:
What? You didn't review the json file properly? 😜

PR #93, opened 2018-11-27
Root search

shredlocker:
# Why?

Add search as part of the root filters.

# What?

- Added search query controller with a search bar. Works similarly to the UISearchController, but wasn't able to use that one...

# Show me

Looks the same, just moved the view.


Sources/Root/Cells/SearchQueryCell.swift
```diff
@@ -4,25 +4,9 @@
...
-        searchBar.delegate = self
-        return searchBar
-    }()
-
-    private var hasTappedClearButton = false
-
-    override var textLabel: UILabel? {
-        return nil
+    var searchBar: UISearchBar? {
+        didSet { didSet(searchBar) }
```
osanoj: I think I would prefer this line to have some newlines 🙂 

And a different name for `didSet` (suggestion: `setupSearchBar`?)

shredlocker: 👍

Sources/Root/FilterRootViewController.swift
```diff
@@ -418,24 +426,28 @@ extension FilterRootViewController: FilterCellDelegate {
...
+extension FilterRootViewController: SearchViewControllerDelegate {
+    public func presentSearchViewController(_ searchViewController: SearchQueryViewController) {
+        searchViewController.searchQuerySuggestionDataSource = searchQuerySuggestionDataSource
+        view.addSubview(searchViewController.view)
+        searchViewController.view.fillInSuperview()
+        view.layoutIfNeeded()
+        // Expand bottom sheet if needed
+        guard let presentationController = navigationController?.presentationController as? BottomSheetPresentationController else { return }
+        searchViewController.previousSizeMode = presentationController.currentContentSizeMode
+        guard presentationController.currentContentSizeMode == .compact else { return }
```
osanoj: It is much nicer if the else is on a newline, makes it easier to read (according to me at least), but primarily it is much easier to debug the code later since you can put a breakpoint there to see that you are exiting early. 

That ^ goes for the rest of the guard-else one-liners too 😄

shredlocker: I'll fix it ;)

Sources/Root/SearchQueryViewController.swift
```diff
@@ -0,0 +1,174 @@
...
+        cell.suggestionLabel.text = suggestion
+        return cell
+    }
+}
+
+// MARK: - TableView Delegate
+
+extension SearchQueryViewController: UITableViewDelegate {
+    public func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let query = suggestions[indexPath.row]
```
osanoj: Its always nice to use the safe way `suggestions[safe: indexPath.row]`

Sources/Root/SearchQueryViewController.swift
```diff
@@ -0,0 +1,194 @@
...
+            searchBar.text = nil
+            searchBar.setShowsCancelButton(false, animated: false)
+            suggestions(forSearchText: nil)
+        }
+        returnToSuperView()
+    }
+
+    public func searchBar(_ searchBar: UISearchBar, textDidChange searchText: String) {
+        // If not active, the user clicked the x-button while not editing and the search should be cancelled
+        if !isActive, searchText.isEmpty {
```
osanoj: I _think_ you might be able to use `searchBar.isDescendant(of: view)` instead of isActive, because when the searchBar is still inside the root table view cell that should return false

shredlocker: Nice one

Sources/Root/FilterRootViewController.swift
```diff
@@ -298,15 +311,12 @@ extension FilterRootViewController: UITableViewDataSource {
...
-            } else {
-                cell.searchText = nil
+            guard let searchQueryFilter = searchQueryFilter else {
+                return searchQueryCell
             }
-            cell.placeholderText = searchQueryFilter?.placeholderText
-            cell.delegate = self
-            return cell
+            let value = selectionDataSource?.value(for: searchQueryFilter)?.first
+            searchQueryCell.searchBar?.text = "Test" // value
```
osanoj: Debug code? 😄

shredlocker: Yep... :P

shredlocker:
It will ;)

PR #100, opened 2018-12-11
Mapfilter part 1 - Distance slider

osanoj:
# Why?
Splitting map filter task into several PRs. This has main focus on making a new slider version. UI should be very similar to `SteppedSlider` that was made for range filters, but this does not require that value increase between steps are equal. 

# What?
This slider takes an array with all the steps that can be "stepped" between when sliding. That way the slider component itself uses the array index as its underlaying value. When handle values that does not come from moving the slider (like initialization values) a value that is not equal to a defined step will put the slider handle between the 2 closest steps.

Hopefully this can later be used for ranges too so that we, for instance, can support price increases of 100 kr per step on left side and 10000 kr on right side of slider. 

# Show me
(The rest of the view is placeholders)
simulator screen shot - iphone 6s plus - 2018-12-11 at 15 23 42


Sources/MapFilter/ValueSliderControl.swift
```diff
@@ -0,0 +1,305 @@
...
+
+import UIKit
+
+typealias SliderValueKind = Comparable & Numeric
+
+protocol ValueSliderControlDelegate: AnyObject {
+    func valueSliderControl<ValueKind: SliderValueKind>(_ valueSliderControl: ValueSliderControl<ValueKind>, didChangeValue: StepValue<ValueKind>)
+}
+
+class ValueSliderControl<ValueKind: SliderValueKind>: UIControl {
```
shredlocker: This doesn't have to be a UIControl does it?

Sources/MapFilter/StepSlider.swift
```diff
@@ -0,0 +1,120 @@
...
+//
+
+import UIKit
+
+protocol StepSliderDelegate: AnyObject {
+    func stepSlider<StepValueKind>(_ stepSlider: StepSlider<StepValueKind>, didChangeValue value: Float)
+    func stepSlider<StepValueKind>(_ stepSlider: StepSlider<StepValueKind>, didChangeRoundedStepValue value: StepValueKind)
+}
+
+class StepSlider<StepValueKind: Comparable & SliderReferenceValue>: UISlider {
```
shredlocker: I would probably put the UISlider directly in the ValueSlider class. Should probably make a slider component together with the range filter thingy

Sources/MapFilter/StepValue.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+struct StepValue<ValueKind: Comparable>: Equatable, Comparable, SliderReferenceValue {
```
shredlocker: Since we have the referenceIndex array as an argument I would probably use that directly instead of using this structure. It is easier to see which values are referenced etc.

Sources/MapFilter/MapFilterView.swift
```diff
@@ -0,0 +1,75 @@
...
+    private lazy var mapView: UIView = {
+        let view = UIView(frame: .zero)
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.backgroundColor = .ice
+        return view
+    }()
+
+    private lazy var distanceSlider: ValueSliderWithLabelView<Int> = {
+        let meterStepValues = [200, 300, 400, 500, 700, 1000, 1500, 2000, 5000, 10000, 20000, 30000, 50000, 75000, 100_000]
+        let referenceIndexes = [1, Int(meterStepValues.count / 2), meterStepValues.count - 2]
```
shredlocker: Instead of using indexes, maybe use relative position like [0, 0.5, 1]. Then you could use that to position the label as well.

osanoj: But I'm pretty sure we want it to point exactly to one of the steps above, and one might not exist for 0.5. I was thinking of using the actual (meter) value, but that has the same issue, it doesn't guarantee that it points to one of the defined steps

Sources/MapFilter/StepSlider.swift
osanoj: I'm not sure it would be possible to handle `accessibilityDecrement` and `accessibilityIncrement` then though. 

The ValueSliderControl would need to be made into something that can be both range and single slider then. Not sure how complex it would make that class. Perhaps better to look into that when this slider is used for ranges too?

Sources/MapFilter/StepValue.swift
osanoj: Hmm, yes maybe, since I had to make a "formatter" class to pass around anyway in the end it might be possible to skip this class. Will look into it 👍

Sources/MapFilter/StepSlider.swift
shredlocker: Jeje, not in this PR ;)

Sources/MapFilter/ValueSliderControl.swift
osanoj: Changed to a view 👍

Sources/MapFilter/StepValue.swift
osanoj: I have removed `StepValue` now 👍

Sources/MapFilter/ValueSliderView.swift
```diff
@@ -0,0 +1,321 @@
...
+            referenceValuesContainer.bottomAnchor.constraint(equalTo: bottomAnchor),
+        ] + referenceViewsConstraints)
+
+        accessibilityElements = [valueSlider]
+        updateAccesibilityValues()
+    }
+
+    func setupReferenceValueView() -> [NSLayoutConstraint] {
+        var constraints = [NSLayoutConstraint]()
+        referenceValueViews = range.enumerated().compactMap { (index, element) -> SliderReferenceValueView<ValueKind>? in
```
shredlocker: Switch this around? 

`let view = referenceValueIndexes.map { index in`
`      let element = range[index].element`
`      return SliderReference....``
`}`

osanoj: 👍

PR #103, opened 2019-01-07
Added podspec

osanoj:
# Why?
We are switching to cocoa pods for dependencies.

# What?
Initial podspec file.


ag12:
A library that simplifies the creation of modern filtering experiences! 😍

PR #108, opened 2019-01-14
Fix showResultsButtonView "jumping" when animating/sliding the bottom sheet

vadymmarkov:
# Why?

`showResultsButtonView` was "jumping" when animating/sliding the bottom sheet because of updates to the safe area.

# What?

Use `safeAreaInsets` of the window to set the bottom anchor of `showResultsButtonView`. Here we access app delegate through `UIApplication.shared.delegate`, so it's not the most elegant solution. But it seems like the only way to get safe area insets before we set constraints. I also tried to use `viewDidLayoutSubviews` or `viewSafeAreaInsetsDidChange`, but it doesn't fix initial "jumping" when we open the bottom sheet with animation.

`UIApplication.shared.delegate?.window??.safeAreaInsets` could also be injected to `FilterRootViewController`, but it seems like overkill for this UI fix.

# Show me

### Before

filter_before

### After

filter_after

### DIFF
--- a/Sources/Root/FilterRootViewController.swift
+++ b/Sources/Root/FilterRootViewController.swift
@@ -134,6 +134,14 @@ public class FilterRootViewController: UIViewController {
         return cell
     }()
 
+    private var bottomLayoutConstant: CGFloat {
+        if #available(iOS 11.0, *) {
+            return UIApplication.shared.delegate?.window??.safeAreaInsets.bottom ?? 0
+        } else {
+            return 0
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


osanoj:
A little late, but it isn't possible to fix this in the bottomsheet controller instead? Since it is because the FIlterRootViewController is inside a bottomsheet that we get this behavior. And after we removed the bottomsheet from this project the filter should be ignorant of how it is presented. 

I also assume this should be a general problem whenever we want to lock something to the bottom safe area of the bottomsheet.

vadymmarkov:
You're right @osanoj, it makes more sense to fix it in FinniversKit 👍

PR #110, opened 2019-01-15
Cleanup in filter selection

osanoj:
# Why?
When integrating with main app there was a lot of unnecessary searches being performed, this was due to excessive calls to inform that selection had changed.

# What?
Separates the internal representation of filter selection and the external api for selection data source. This should remove a lot of unnecessary calls to ParameterBasedFilterInfoSelectionDataSource delegate, there should max be 1 call for each function call in FilterSelectionDataSource-implementation.

### DIFF
--- a/Sources/FINNFilterImplementation/FINNFilterSelectionData.swift
+++ b/Sources/FINNFilterImplementation/FINNFilterSelectionData.swift
@@ -1,0 +1,219 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+class FINNFilterSelectionData: NSObject {
+    private struct GeoKey {
+        static let latitude = "lat"
+        static let longitude = "lon"
+        static let radius = "radius"
+        static let locationName = "geoLocationName"
+    }
+
+    private(set) var selectionValues: [String: [String]]
+
+    init(selectionValues: [String: [String]]) {
+        self.selectionValues = selectionValues
+    }
+
+    convenience override init() {
+        self.init(selectionValues: [:])
+    }
+}
+
+extension FINNFilterSelectionData {
+    func setSelectionValues(_ values: [String], for key: String) {
+        selectionValues[key] = values
+    }
+
+    func setSelectionValue(_ value: String, for key: String) {
+        setSelectionValues([value], for: key)
+    }
+
+    func addSelectionValue(_ value: String, for key: String) {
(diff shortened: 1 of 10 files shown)


Sources/FINNFilterImplementation/FINNFilterSelectionData.swift
```diff
@@ -0,0 +1,219 @@
...
+        if let previousValues = selectionValues[key] {
+            setSelectionValues(previousValues.filter({ $0 != value }), for: key)
+        }
+    }
+
+    func removeSelectionValues(_ key: String) {
+        selectionValues.removeValue(forKey: key)
+    }
+
+    func selectionValues(for name: String) -> [String] {
```
vadymmarkov: Should it be `key` instead of `name`? Just because it's called `key` in other methods.

Sources/FINNFilterImplementation/FINNFilterSelectionData.swift
```diff
@@ -0,0 +1,219 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+class FINNFilterSelectionData: NSObject {
```
vadymmarkov: Does it need to be `NSObject`?

Sources/FINNFilterImplementation/FINNFilterSelectionData.swift
```diff
@@ -0,0 +1,219 @@
...
+        self.selectionValues = selectionValues
+    }
+
+    convenience override init() {
+        self.init(selectionValues: [:])
+    }
+}
+
+extension FINNFilterSelectionData {
+    func setSelectionValues(_ values: [String], for key: String) {
```
vadymmarkov: Do we have tests for methods in this class?

Sources/FINNFilterImplementation/FINNFilterSelectionData.swift
osanoj: That should not be needed 👍

Sources/FINNFilterImplementation/FINNFilterSelectionData.swift
osanoj: Indirectly, there are tests for the public api implemented in ParameterBasedFilterInfoSelectionDataSource, I will have a look at if it makes sense to add test to this class too 🤔

PR #115, opened 2019-01-17
Refactor constraints in stepper filter view and view controller

shredlocker:
# Why?

Stepper filter was not behaving as expected.

# What?

Change constraints to make stepper filter view attached to bottom sheet.

# Show me

### Before

stepper-filter

### After

stepper-filter-fix


Sources/StepperFilter/StepperFilterViewController.swift
```diff
@@ -35,7 +35,11 @@ public class StepperFilterViewController: UIViewController, FilterContainerViewC
     public override func viewDidLoad() {
         super.viewDidLoad()
         view.addSubview(stepperFilterView)
-        stepperFilterView.fillInSuperview()
+        NSLayoutConstraint.activate([
+            stepperFilterView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            stepperFilterView.topAnchor.constraint(equalTo: view.topAnchor, constant: 148),
```
osanoj: I don't like that `148` magic value... does it work on iPhone 4S? 😛 

Is there no other way to position it without magic constants? Or calculate it based on height?

shredlocker: Haha, I'll figure something out! 😄

osanoj: `.veryLargeSpacing * 2`  😜

shredlocker: Looks like that works well 😄

PR #116, opened 2019-01-18
Merge range sliders

vadymmarkov:
# Why?

Because we want to use new "step slider" for ranges and work with arrays of values instead of steps.

# What

- The slider used in ranges was replaced by the one used by map filter
- Refactor range filters using new `StepSliderInfo` type, which holds all necessary data to setup and manage sliders
- `RangeFilterInfoBuilder` still creates ranges with increments and equal steps as before. In order to take advantage of `StepSlider` we should replace that with arrays of values
- Make it possible to set lower and upper bound offsets in `StepSlider` 
- Hopefully didn't break anything 🤞😄 

# Show me

### Before

before1

### After

after1


bstien:
I tried entering some other value outside the ones in the list😅

charcoal-rangefilter

vadymmarkov:
Good catch 😄

vadymmarkov:
It should be fixed now:

after22

PR #118, opened 2019-01-21
Configure market MC

bstien:
# Why?
Users should be able to filter for MC.

# What?
- Implemented MC in `FilterKey` & `FilterMarket`.
- Added Demo.

# Show me
### MC
screen shot 2019-01-21 at 13 24 03
screen shot 2019-01-21 at 13 24 15

### Scooter & moped
screen shot 2019-01-21 at 13 26 01

### Snowmobile
screen shot 2019-01-21 at 13 26 04

### ATV
screen shot 2019-01-21 at 13 26 09


Demo/Resources/snowmobile.json
```diff
@@ -0,0 +1,348 @@
+{
+    "market": "snowmobile",
```
vadymmarkov: Where do you get the from? Just curious.

Sources/FINNFilterImplementation/FilterBuilder/RangeFilterInfoBuilder.swift
```diff
@@ -219,4 +221,76 @@ private extension RangeFilterInfoBuilder {
...
+            rangeBoundsOffsets = (hasLowerBoundOffset: true, hasUpperBoundOffset: true)
+            increment = 1
+            accessibilityValues = (stepIncrement: nil, valueSuffix: nil)
+            appearanceProperties = (usesSmallNumberInputFont: false, displaysUnitInNumberInput: false, isCurrencyValueRange: false)
+        case .engineEffect:
+            lowValue = 0
+            highValue = 200
+            unit = "hk"
+            rangeBoundsOffsets = (hasLowerBoundOffset: false, hasUpperBoundOffset: true)
+            increment = 10
```
vadymmarkov: I think we can start using arrays of values instead of increment. Not sure what they should be, but that's the goal I think.

Demo/Resources/snowmobile.json
bstien: From this subcategory. I based it on the URL we're calling once it's selected.

screen shot 2019-01-21 at 14 17 07

bstien: Or did you mean the data itself?

vadymmarkov: Yeah, the data 😄

Sources/FINNFilterImplementation/FilterBuilder/RangeFilterInfoBuilder.swift
bstien: I agree 👍

PR #119, opened 2019-01-21
Fix range sliders crossing over each other

vadymmarkov:
# Why?

Because it's confusing for the user when range sliders overlap or cross over each other.

# What?

- Make left and right slider stick on the same value when they collide
- Change positions of reference values so it's clear for the user why sliders don't overlap
- Add input validations when the "from" value goes over the "to" value and the other way around

# Show me

### Before

recordbefore

### After

record


shredlocker:
The collision haptics are bit crazy :P I guess there should only be one haptic response the moment they collide, but there is a haptic whenever I move my thumb during collision.

vadymmarkov:
The collision and input validation haptics are fixed now. Also, both number input and unit label are highlighted in red if validation fails:

simulator screen shot - iphone xs - 2019-01-22 at 14 33 50

PR #120, opened 2019-01-22
Change constraints in stepper demo view

shredlocker:
# Why?

Stepper filter demo was broken after changing the constraints in StepperFilterView.

# What?

Fixes the stepper filter demo view constraints so it works with updated constraints.

### DIFF
--- a/Demo/StepperFilterDemoView.swift
+++ b/Demo/StepperFilterDemoView.swift
@@ -23,7 +23,11 @@ class StepperFilterDemoView: UIView {
     override init(frame: CGRect) {
         super.init(frame: frame)
         addSubview(stepperFilterView)
-        stepperFilterView.fillInSuperview()
+        NSLayoutConstraint.activate([
+            stepperFilterView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            stepperFilterView.trailingAnchor.constraint(equalTo: trailingAnchor),
+            stepperFilterView.bottomAnchor.constraint(equalTo: topAnchor, constant: 510 / 2),
+        ])
     }
 
     required init?(coder aDecoder: NSCoder) {
--- a/Demo/StepperFilterDemoView.swift
+++ b/Demo/StepperFilterDemoView.swift
@@ -13,6 +13,9 @@ struct StepperData: StepperFilterInfoType {
 }
 
 class StepperFilterDemoView: UIView {
+    private var didSetConstant = false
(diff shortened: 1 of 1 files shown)


Demo/StepperFilterDemoView.swift
```diff
@@ -23,7 +23,11 @@ class StepperFilterDemoView: UIView {
     override init(frame: CGRect) {
         super.init(frame: frame)
         addSubview(stepperFilterView)
-        stepperFilterView.fillInSuperview()
+        NSLayoutConstraint.activate([
+            stepperFilterView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            stepperFilterView.trailingAnchor.constraint(equalTo: trailingAnchor),
+            stepperFilterView.bottomAnchor.constraint(equalTo: topAnchor, constant: 510 / 2),
```
vadymmarkov: Magic number? 😄

Demo/StepperFilterDemoView.swift
```diff
@@ -23,7 +23,11 @@ class StepperFilterDemoView: UIView {
     override init(frame: CGRect) {
         super.init(frame: frame)
         addSubview(stepperFilterView)
-        stepperFilterView.fillInSuperview()
+        NSLayoutConstraint.activate([
+            stepperFilterView.leadingAnchor.constraint(equalTo: leadingAnchor),
```
vadymmarkov: Don't you need either `topAnchor` or `heightAnchor` here?

Demo/StepperFilterDemoView.swift
shredlocker: The idea was that the bottom sheet would have a height of at least 510. 
Any suggestions on the best way to get the view's height?

shredlocker: Added a fix, but if you have a better way let me know! 😄

vadymmarkov:
Isn't that what you want to achieve?
```swift
NSLayoutConstraint.activate([
    stepperFilterView.topAnchor.constraint(equalTo: centerYAnchor),
    stepperFilterView.leadingAnchor.constraint(equalTo: leadingAnchor),
    stepperFilterView.trailingAnchor.constraint(equalTo: trailingAnchor)
 ])
```

shredlocker:
This creates the original problem. I want it to be a fixed constant from the top, and the constant should be half the height of bottom sheet

vadymmarkov:
Ok, I see now. Seeing that the height of bottom sheet is dynamic then your last fix should probably work 😉

PR #121, opened 2019-01-22
Refactor/filter markets

bstien:
# Why?
The different markets were starting to grow quite unmageable, seeing as they were all placed into a single file/`enum`.

⚠️ I'd like to get some feedback on naming of the different markets. As of now, they have names like `FilterMarketBap`, `FilterMarketRealestate` and `FilterMarketMC`.

# What?
- Pulled out filterConfig for each market into a separate file.
- Sorted and removed newlines from `FilterKey`.
- The responsibility for creating rangeFilters are now placed within each filterMarket.
- Pulled out protocol `FilterConfiguration` into a separate file.
- Removed unused code after refactoring.


bstien:
I branched out from my branch in #118 and started this work before that was merged. It seems my commits from that PR is included in the changes for this one.

Not sure if that makes any difference, though

Sources/FINNFilterImplementation/FilterMarkets/FilterConfiguration.swift
```diff
@@ -0,0 +1,13 @@
...
+//
+
+import Foundation
+
+protocol FilterConfiguration {
+    func handlesVerticalId(_ vertical: String) -> Bool
+    var preferenceFilterKeys: [FilterKey] { get }
+    var supportedFiltersKeys: [FilterKey] { get }
+    var mapFilterKey: FilterKey? { get }
+    func createFilterInfoFrom(filterData: FilterData) -> FilterInfoType?
```
osanoj: Maybe change it to something like `func createFilterInfoFrom(rangeFilterData: FilterData) -> FilterInfoType?` so it is clearer that it is only for filter data that has `isRange == true`

bstien: Makes sense, I'll update it 👍

PR #122, opened 2019-01-22
Mapfilter part 5 and other updates for main app integration

osanoj:
# Why?
There were several issues when integrating with main app.

# What?
- Some changes to map filters that were needed when integrating Charcoal with our main app.
- Also added an `applyButtonTapped` function to `FilterViewController` to make it easy for subclasses to apply changes when apply button is tapped. Used by `MapFilterViewController`.
- Fixed issue in FilterRootViewController where filter data changes (after a search) were incorrectly marked as needing to reload UI.

### DIFF
--- a/Sources/MapFilter/MapFilterViewController.swift
+++ b/Sources/MapFilter/MapFilterViewController.swift
@@ -44,9 +44,7 @@ public class MapFilterViewController: FilterViewController {
         showApplyButton(true, animated: false)
     }
 
-    public override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
-        guard let currentSelection = selectionDataSource.geoValue(for: filterInfo) else {
+        guard let currentSelection = selectionDataSource.geoValue() else {
             return
         }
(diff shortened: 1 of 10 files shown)


Sources/MapFilter/MapFilterViewController.swift
```diff
@@ -41,25 +41,7 @@ public class MapFilterViewController: FilterViewController {
...
-
-    public override func viewWillDisappear(_ animated: Bool) {
-        super.viewWillDisappear(animated)
-
-        // TODO: This should be done when apply or back is pressed
-        guard let mapFilterView = mapFilterView, let coordinate = mapFilterViewManager?.centerCoordinate else {
-            return
-        }
-        selectionDataSource.setValue(latitude: coordinate.latitude, longitude: coordinate.longitude, radius: mapFilterView.currentRadius, locationName: nil, for: filterInfo)
+        showApplyButton(true)
```
shredlocker: You might want to use `showApplyButton(true, animated: false)` as before.

osanoj: Ah, I see now that it defaults to `animated = true`

PR #129, opened 2019-01-28
Refactor range slider

vadymmarkov:
# Why?

To simplify the implementation of range slider and make it easier to handle corner cases.

# What?

- Use `Int` instead of generics types for step sliders
- Introduce new `Step` enum for setting values on `StepSlider`
- Clean up the code
- Improve input validations (again)

Please test and let me know if there any issues with number input left. As for me it seems unbreakable now 😅 

# Show me

No UI changes.


vadymmarkov:
@bstien The bug you've found should bee fixed now 😉

PR #131, opened 2019-01-30
Introduce context filters

shredlocker:
# Why?

We need a way to define context aware filters and a way to display these in the root filter.

# What?

- Added function for defining context aware filters in each market. Each supported filter key can define a list of filter keys which can be used as context filters.
- Added dot to FilterCell which is hidden by default and will be shown if it is context aware
- Added property to FilterInfoType to specify if it is a context aware filter info
- Two small json filter files, one with a context filter and one without.
- Demo with the small filter.

# Show me

simulator screen shot - iphone xs - 2019-01-30 at 15 11 57

### DIFF
--- a/Sources/Dependencies/FilterDependencyContainer.swift
+++ b/Sources/Dependencies/FilterDependencyContainer.swift
@@ -5,8 +5,6 @@
 import Foundation
 
 public class FilterDependencyContainer {
-    public weak var applyButtonDelegate: ApplySelectionButtonDelegate?
-
     private let selectionDataSource: FilterSelectionDataSource
     private let searchQuerySuggestionsDataSource: SearchQuerySuggestionsDataSource?
     private weak var filterRootStateControllerDelegate: FilterRootStateControllerDelegate?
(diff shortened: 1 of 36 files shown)


shredlocker:
I made it work, but I'm not really happy about the way I made it work with the applySelectionButtonDelegate :P could use some clever ideas on that part.

Sources/Root/Cells/FilterCell.swift
```diff
@@ -57,13 +80,16 @@ class FilterCell: UITableViewCell {
         super.prepareForReuse()
         selectedValues = nil
         filterName = nil
+        dotView.isHidden = true
```
vadymmarkov: You can also do `isContextFilter = false` instead.

shredlocker: Aha, that is true! 😄

Sources/Dependencies/FilterDependencyContainer.swift
```diff
@@ -5,6 +5,8 @@
 import Foundation
 
 public class FilterDependencyContainer {
+    public weak var applyButtonDelegate: ApplySelectionButtonDelegate?
```
vadymmarkov: Maybe it's better to pass it as a parameter in `makeMultiLevelListSelectionFilterViewController` since it's the only place it's used.

shredlocker: You are right, you could even set the delegate in the `navigate(to destination:)` function instead of passing it.

vadymmarkov:
Why do you need this `applyButtonDelegate` btw?

osanoj:
Yes, that should not be done from inside the filters, giving new data to rootStateController should be from "outside"

osanoj:
Maybe you can try using a custom `ContextFilterDemoSelectionDatasource`, that can trigger loading new data into the `rootStateController`. And remove the ApplySelectionButtonDelegate.

osanoj:
When used in the real application setting new filter data will of course not be handled by the selection datasource, instead when the search results response is retrieved and parsed. But since we don't have searches in the demo you need to make a little workaround to simulate it

Sources/FINNFilterImplementation/FilterMarkets/FilterMarketBap.swift
```diff
@@ -31,6 +31,15 @@ extension FilterMarketBap: FilterConfiguration {
         return .location
     }
 
+    func contextFilterKeys(for key: FilterKey) -> [FilterKey] {
+        switch key {
+        case .category:
```
osanoj: Why are you grouping the context filters to another filter? This isn't really how they work...

shredlocker:
Going to make a new PR

PR #132, opened 2019-02-01
Prevent panGesture in rangeFilters view

bstien:
# Why?
It was possible to make the bottomSheet expand/collapse by swiping within the view of a rangeFilter. The bottomSheet should only do this when sliding the top section.

# What?
- Swallow panGestures and prevent them from propagating upwards in the view hierarchy.

# Show me
### Before & After
I made the backgroundColor `lightGray` just to make it visible where I'm swiping in the view.

range_filter_pan_before
range_filter_pan_after

### DIFF
--- a/Sources/RangeFilter/RangeFilterViewController.swift
+++ b/Sources/RangeFilter/RangeFilterViewController.swift
@@ -5,8 +5,16 @@
 import Foundation
 
 public final class RangeFilterViewController: FilterViewController {
+
+    // MARK: - Private attributes
+
     private let filterInfo: RangeFilterInfoType
     private var currentRangeValue: RangeValue?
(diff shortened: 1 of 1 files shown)


osanoj:
Is it possible to do this at bottomsheet level instead? Because Charcoal should not really care/know it is being presented in one. 

And we want this for all views inside the bottomsheet.

bstien:
You're absolutely right, it should be done there instead. I'll leave this open until I have some progress in `FinniversKit`.

shredlocker:
Ww don't want this for all view in the bottomshett. Only in the filters.

bstien:
Opened a PR in `FinniversKit` fixing this issue. Closing this one.

PR #135, opened 2019-02-04
Remove Carthage, use cocoapods

bstien:
# Why?
> ✨🌙 _\*Tumbleweed rolls by, crickets chirping\*_ 
> _Cocoapods:_ This project isn't big enough for the two of us
> _Carthage:_ What do you mea....
> _Cocoapods:_ **BANG** 🔥🔫🤠

# What?
- Created Podfile with the few dependencies we have.
- Removed all traces of Carthage.


vadymmarkov:
Usually it's a good practise for iOS open source libraries to support both CocoaPods and Carthage. Note that It won't be possible to install Charcoal via Carthage if we merge this.

osanoj:
Good point, maybe we should try to support both? We will need to remember to maintain both cartfile and podfile though. Anything else we would need to deal with that might make it more complicated for us?

vadymmarkov:
We already have a project configured for using Carthage, and the Cartfile is pretty simple with only FinniversKit and HockeySDK included so far. The only thing we need to maintain is Podspec, but we'll have to do it regardless. So I think we should be good.

vadymmarkov:
So should we close this one and keep Carthage and CocoaPods support?

osanoj:
Yes, let's close it

PR #139, opened 2019-02-08
Filter tree structure

shredlocker:
# Why?

This PR aims to simplify the logic behind the filter selection process. Instead of having two different objects handling UI state and filter selection, I have made a structure that will handle both. The goal
was to make the filter state more connected to the UI.

The previous selection process
```swift
selectionDataSource.addValue(value, for: filterInfo)
// and
selectionDataSource.clearValue(value, for: filterInfo)
```
is replaced with
``` swift
filterNode.isSelected = true // or false
```
where each filter node represents a single url query item.

Each view controller will receive a filter node during initialisation and it will be responsible for handling the child nodes of that filter node. F.eks:

```swift
func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
    return filterNode.children.count
}

func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
    let cell = tableView.dequeue(CCListFilterCell.self, for: indexPath)
    cell.configure(for: filterNode.child(at: indexPath.row))
    return cell
}
```

When the user have applied changes to the filter selection, the updated selection can easily be retrieved like so:
```swift
let queryString = filter.urlEncoded
```

I believe that this tree structure will help improve the navigation and flow of data.

# What?

- Added a Filter structure which is composed of multiple FilterNodes. Every filter node can act as a key-value pair and represent a single url query item or it can have child nodes and represent a group of key-value pairs.
- Added RangeFilterNode and MapFilterNode which are special cases where the child nodes are hard coded for convenience.
- Added implementation to demonstrate how you can build the filter tree using the FilterSetup object.
- Added tests for filter and filter nodes.

# Show me

No UI

### DIFF
--- a/Sources/CCFilter/CCFilter.swift
+++ b/Sources/CCFilter/CCFilter.swift
@@ -5,7 +5,12 @@
 import Foundation
 
 public class CCFilter {
+
+    // MARK: - Private properties
+
     private var root: CCFilterNode
+
+    // MARK: - Setup
 
     public init(root: CCFilterNode) {
         self.root = root
@@ -13,8 +18,8 @@ public class CCFilter {
 }
 
 public extension CCFilter {
-    var urlEncoded: String {
-        return root.urlItems.joined(separator: "&")
+    var queryItems: [URLQueryItem] {
+        return root.queryItems
     }
 
     func reset() {
--- a/Sources/CCFilter/CCFilter.swift
+++ b/Sources/CCFilter/CCFilter.swift
@@ -1,0 +1,23 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+public class CCFilter {
+    private var root: CCFilterNode
+
+    public init(root: CCFilterNode) {
+        self.root = root
+    }
+}
+
+public extension CCFilter {
+    var urlEncoded: String {
+        return root.urlItems.joined(separator: "&")
+    }
+
+    func reset() {
+        root.reset()
+    }
+}
--- a/Sources/CCFilterNode/CCFilterNode.swift
+++ b/Sources/CCFilterNode/CCFilterNode.swift
@@ -1,0 +1,72 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+private protocol CCFilterNodeDelegate: class {
+    func childNodeDidChangeSelection(_ childNode: CCFilterNode)
+}
+
+public class CCFilterNode {
+
+    // MARK: - Public properties
+
+    public let title: String
+    public let name: String
+    public var value: String?
+    public var numberOfResults: Int
+
+    public var isSelected: Bool {
+        didSet { delegate?.childNodeDidChangeSelection(self) }
+    }
+
+    // MARK: - Private properties
+
+    private(set) var children: [CCFilterNode] = []
+    private weak var delegate: CCFilterNodeDelegate?
+
+    // MARK: - Setup
+
+    public init(title: String, name: String, value: String? = nil, isSelected: Bool = false, numberOfResults: Int = 0) {
+        self.title = title
+        self.name = name
+        self.value = value
+        self.isSelected = isSelected
+        self.numberOfResults = numberOfResults
+    }
+
+    // MARK: - Public methods
+
+    func add(child: CCFilterNode, at index: Int? = nil) {
+        if let index = index { children.insert(child, at: index) }
+        else { children.append(child) }
+        child.delegate = self
+    }
+
+    func child(at index: Int) -> CCFilterNode {
+        return children[index]
+    }
+
+    func reset() {
+        isSelected = false
+        children.forEach { $0.reset() }
+    }
+
+    var isLeafNode: Bool {
+        return children.count == 0
+    }
+
+    var urlItems: [String] {
+        if isSelected, let value = value {
+            return ["\(name)=\(value)"]
+        }
+        return children.reduce([]) { $0 + $1.urlItems }
+    }
+}
+
+extension CCFilterNode: CCFilterNodeDelegate {
+    func childNodeDidChangeSelection(_ childNode: CCFilterNode) {
+        isSelected = children.reduce(true) { $0 && $1.isSelected }
+    }
+}
--- a/Sources/CCFilterNode/CCFilterNode.swift
+++ b/Sources/CCFilterNode/CCFilterNode.swift
@@ -15,7 +15,7 @@ public class CCFilterNode {
     public let title: String
     public let name: String
     public var value: String?
-    public var numberOfResults: Int
+    public let numberOfResults: Int
 
     public var isSelected: Bool {
         didSet { delegate?.childNodeDidChangeSelection(self) }
@@ -39,12 +39,16 @@ public class CCFilterNode {
     // MARK: - Public methods
 
     func add(child: CCFilterNode, at index: Int? = nil) {
-        if let index = index { children.insert(child, at: index) }
-        else { children.append(child) }
+        if let index = index {
+            children.insert(child, at: index)
+        } else {
+            children.append(child)
+        }
         child.delegate = self
     }
 
-    func child(at index: Int) -> CCFilterNode {
+    func child(at index: Int) -> CCFilterNode? {
+        guard index < children.count else { return nil }
         return children[index]
     }
 
@@ -54,19 +58,19 @@ public class CCFilterNode {
     }
 
     var isLeafNode: Bool {
-        return children.count == 0
+        return children.isEmpty
     }
 
-    var urlItems: [String] {
+    var queryItems: [URLQueryItem] {
         if isSelected, let value = value {
-            return ["\(name)=\(value)"]
+            return [URLQueryItem(name: name, value: value)]
         }
-        return children.reduce([]) { $0 + $1.urlItems }
+        return children.reduce([]) { $0 + $1.queryItems }
     }
 }
 
 extension CCFilterNode: CCFilterNodeDelegate {
     func childNodeDidChangeSelection(_ childNode: CCFilterNode) {
-        isSelected = children.reduce(true) { $0 && $1.isSelected }
+        isSelected = children.allSatisfy { $0.isSelected }
     }
 }
--- a/Sources/CCFilterNode/CCMapFilterNode.swift
+++ b/Sources/CCFilterNode/CCMapFilterNode.swift
@@ -4,29 +4,36 @@
 
 import Foundation
 
-extension CCMapFilterNode {
-    enum Key: String, CaseIterable {
-        case lat, lon, radius, geoLocationName
-    }
+public class CCMapFilterNode: CCFilterNode {
 
-    enum Index: Int, CaseIterable {
-        case lat, lon, radius, geoLocationName
-    }
-}
+    // MARK: - Public properies
 
-class CCMapFilterNode: CCFilterNode {
-    static let filterKey = "map"
+    public static let filterKey = "map"
 
-    init(title: String, name: String) {
+    // MARK: - Internal properties
+
+    let latitudeNode: CCFilterNode
+    let longitudeNode: CCFilterNode
+    let radiusNode: CCFilterNode
+    let geoLocationNode: CCFilterNode
+
+    // MARK: - Setup
+
+    public init(title: String, name: String) {
+        latitudeNode = CCFilterNode(title: "", name: "lat")
+        longitudeNode = CCFilterNode(title: "", name: "lon")
+        radiusNode = CCFilterNode(title: "", name: "radius")
+        geoLocationNode = CCFilterNode(title: "", name: "geoLocationName")
         super.init(title: title, name: name, value: nil, isSelected: false, numberOfResults: 0)
         setup()
     }
 }
 
-extension CCMapFilterNode {
+private extension CCMapFilterNode {
     func setup() {
-        Index.allCases.forEach {
-            add(child: CCFilterNode(title: "", name: Key.allCases[$0.rawValue].rawValue))
-        }
+        add(child: latitudeNode)
+        add(child: longitudeNode)
+        add(child: radiusNode)
+        add(child: geoLocationNode)
     }
 }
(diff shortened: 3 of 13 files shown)


Sources/CCFilterNode/CCFilterNode.swift
```diff
@@ -0,0 +1,72 @@
...
+        if isSelected, let value = value {
+            return ["\(name)=\(value)"]
+        }
+        return children.reduce([]) { $0 + $1.urlItems }
+    }
+}
+
+extension CCFilterNode: CCFilterNodeDelegate {
+    func childNodeDidChangeSelection(_ childNode: CCFilterNode) {
+        isSelected = children.reduce(true) { $0 && $1.isSelected }
```
bstien: I think this will be more readable:

```swift
isSelected = children.allSatisfy({ $0.isSelected })
```

Sources/FINNFilterImplementation/BackendModel/FilterSetup.swift
```diff
@@ -47,6 +47,31 @@ public struct FilterSetup: Decodable {
...
     }
 
+    public func asCCFilter() -> CCFilter? {
+        guard let filterMarket = FilterMarket(market: market) else {
+            return nil
+        }
+
+        let searchQueryNode = CCFilterNode(title: "Ord i annonsen", name: "q")
+
+        let preferenceFilters = filterMarket.preferenceFilterKeys.compactMap { self.filterData(forKey: $0) }
```
bstien: You may omit `self` here.

Sources/FINNFilterImplementation/BackendModel/FilterSetup.swift
```diff
@@ -47,6 +47,31 @@ public struct FilterSetup: Decodable {
...
+        let preferenceNode = CCFilterNode(title: "Preferences", name: "preferences")
+        preferenceFilters.forEach { preferenceNode.add(child: $0.filterNode()) }
+
+        let filters = filterMarket.supportedFiltersKeys.compactMap { self.filterData(forKey: $0) }
```
bstien: And here

Sources/CCFilter/CCFilter.swift
```diff
@@ -0,0 +1,23 @@
...
+    private var root: CCFilterNode
+
+    public init(root: CCFilterNode) {
+        self.root = root
+    }
+}
+
+public extension CCFilter {
+    var urlEncoded: String {
+        return root.urlItems.joined(separator: "&")
```
bstien: I guess we need to URL encode each of the values, not just join them by `&`.

I.e. passing `Søk & test` as value to a node should return `q=S%C3%B8k%20%26%20test`, but now returns `q=Søk & test`. This will break the search😅

vadymmarkov: Can we perhaps use `URLQueryItem` instead, it will do url encoding for us?

shredlocker: I fix!

Sources/CCFilterNode/CCFilterNode.swift
```diff
@@ -0,0 +1,72 @@
...
+    func childNodeDidChangeSelection(_ childNode: CCFilterNode)
+}
+
+public class CCFilterNode {
+
+    // MARK: - Public properties
+
+    public let title: String
+    public let name: String
+    public var value: String?
```
vadymmarkov: Could `value` and `numberOfResults` be constants?

Sources/CCFilter/CCFilter.swift
bstien: Maybe we could return an array of `URLQueryItem` in `urlItems` instead of strings?

Sources/CCFilterNode/CCFilterNode.swift
```diff
@@ -0,0 +1,72 @@
...
+        self.name = name
+        self.value = value
+        self.isSelected = isSelected
+        self.numberOfResults = numberOfResults
+    }
+
+    // MARK: - Public methods
+
+    func add(child: CCFilterNode, at index: Int? = nil) {
+        if let index = index { children.insert(child, at: index) }
```
vadymmarkov: Not sure I like this `if else` formatting, I'd prefer:
```swift
if {
} else {
}
```

Sources/CCFilterNode/CCFilterNode.swift
```diff
@@ -0,0 +1,72 @@
...
+        else { children.append(child) }
+        child.delegate = self
+    }
+
+    func child(at index: Int) -> CCFilterNode {
```
vadymmarkov: Shouldn't it be optional in the index is out of bounds? You could also implement custom subscript instead of this function.

Sources/CCFilterNode/CCFilterNode.swift
```diff
@@ -0,0 +1,72 @@
...
+        return children[index]
+    }
+
+    func reset() {
+        isSelected = false
+        children.forEach { $0.reset() }
+    }
+
+    var isLeafNode: Bool {
+        return children.count == 0
```
vadymmarkov: Could be `children.isEmpty`

Sources/CCFilterNode/CCMapFilterNode.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+extension CCMapFilterNode {
+    enum Key: String, CaseIterable {
+        case lat, lon, radius, geoLocationName
```
vadymmarkov: Perhaps `latitude` and `longitude` would be more readable?

Sources/CCFilterNode/CCMapFilterNode.swift
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    enum Index: Int, CaseIterable {
+        case lat, lon, radius, geoLocationName
```
vadymmarkov: The same comment as above.

Sources/CCFilterNode/CCMapFilterNode.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+extension CCMapFilterNode {
+    enum Key: String, CaseIterable {
```
vadymmarkov: Could it be private and/or nested in `CCMapFilterNode`?

Sources/CCFilterNode/CCMapFilterNode.swift
```diff
@@ -0,0 +1,32 @@
...
+        case lat, lon, radius, geoLocationName
+    }
+
+    enum Index: Int, CaseIterable {
+        case lat, lon, radius, geoLocationName
+    }
+}
+
+class CCMapFilterNode: CCFilterNode {
```
vadymmarkov: Shouldn't it be public?

Sources/FINNFilterImplementation/BackendModel/FilterData.swift
```diff
@@ -27,6 +27,20 @@ struct FilterData: Decodable {
         queries = try container.decodeIfPresent([FilterDataQuery].self, forKey: CodingKeys.queries) ?? []
     }
 
+    public func filterNode() -> CCFilterNode {
+        if let isRange = isRange, isRange {
+            let rangeNode = CCRangeFilterNode(title: title, name: parameterName)
```
vadymmarkov: Why not just return here?

Sources/FINNFilterImplementation/BackendModel/FilterData.swift
```diff
@@ -27,6 +27,20 @@ struct FilterData: Decodable {
         queries = try container.decodeIfPresent([FilterDataQuery].self, forKey: CodingKeys.queries) ?? []
     }
 
+    public func filterNode() -> CCFilterNode {
+        if let isRange = isRange, isRange {
```
vadymmarkov: You could also do `if isRange == true`.

Sources/FINNFilterImplementation/BackendModel/FilterSetup.swift
```diff
@@ -47,6 +47,31 @@ public struct FilterSetup: Decodable {
         }
     }
 
+    public func asCCFilter() -> CCFilter? {
+        guard let filterMarket = FilterMarket(market: market) else {
+            return nil
+        }
+
+        let searchQueryNode = CCFilterNode(title: "Ord i annonsen", name: "q")
```
vadymmarkov: Should it go to localisable strings?

Sources/FINNFilterImplementation/BackendModel/FilterSetup.swift
```diff
@@ -47,6 +47,31 @@ public struct FilterSetup: Decodable {
...
+
+        let preferenceFilters = filterMarket.preferenceFilterKeys.compactMap { self.filterData(forKey: $0) }
+        let preferenceNode = CCFilterNode(title: "Preferences", name: "preferences")
```
vadymmarkov: Localizable.strings?

Sources/FINNFilterImplementation/BackendModel/FilterSetup.swift
```diff
@@ -47,6 +47,31 @@ public struct FilterSetup: Decodable {
...
+        preferenceFilters.forEach { preferenceNode.add(child: $0.filterNode()) }
+
+        let filters = filterMarket.supportedFiltersKeys.compactMap { self.filterData(forKey: $0) }
+        let filterNodes = filters.map { $0.filterNode() }
+
+        if let locationNode = filterNodes.first(where: { $0.name == FilterKey.location.rawValue }) {
+            let mapNode = CCMapFilterNode(title: "Område i kart", name: CCMapFilterNode.filterKey)
```
vadymmarkov: Same comment about strings.

Sources/CCFilter/CCFilter.swift
osanoj: Yes, much better to use `URLQueryItem` 👍

Sources/CCFilterNode/CCMapFilterNode.swift
osanoj: Agree, better to use `latitude` and `longitude`. And newlines after each case (a general recommendation, not just for this enum)

Sources/CCFilterNode/CCFilterNode.swift
shredlocker: I think the view controller should make sure the index is valid before calling this function. I simply use this function for consistency, I have the function to add a child so I think it is nice to have a function to get a child as well :)

Sources/CCFilterNode/CCFilterNode.swift
shredlocker: :P I'll change it

Sources/CCFilterNode/CCFilterNode.swift
shredlocker: The value will change in the range filters, but I think number of results could be constant.

Sources/CCFilterNode/CCFilterNode.swift
shredlocker: That's better! 😄

Sources/CCFilterNode/CCMapFilterNode.swift
shredlocker: Agreed!

Sources/CCFilterNode/CCMapFilterNode.swift
shredlocker: Yes, it should!

Sources/FINNFilterImplementation/BackendModel/FilterSetup.swift
shredlocker: I'll fix the strings

Sources/FINNFilterImplementation/BackendModel/FilterData.swift
shredlocker: I had a line in between at some point, I'll fix this

Sources/CCFilterNode/CCFilterNode.swift
vadymmarkov: I think it's easy to forget this check in the view controller and get a crash.

Sources/FINNFilterImplementation/BackendModel/FilterData.swift
shredlocker: Oo, nice! :)

Sources/CCFilterNode/CCMapFilterNode.swift
shredlocker: Could potentially remove these and have properties for each node.

osanoj:
Are these nodes supposed to replace both the selection datasource and the `FilterInfo` protocols and classes? How would you handled the updates after a search? You will get new values for `numberOfResults` and some items might be missing (but should still be shown) and some new items might be added (context filters)

Sources/CCFilterNode/CCFilterNode.swift
osanoj: Yes, much better to assume the caller might use a bad index. We have the custom `[safe: index]` subscript that is always good to use

Sources/CCFilterNode/CCFilterNode.swift
osanoj: So `value` will represent the values retrieved from backend (filter setup) and for some types the current selection for that node? Could be slightly confusing 🤔

shredlocker:
This is a dynamic data structure so it is no problem adding context filters and changing values when we get new filters with the search.

Sources/CCFilterNode/CCFilterNode.swift
shredlocker: It represents the same thing, except for most filters the value is static and for others it is dynamic.

osanoj: Hmm, yes I guess that is true. The value to use when filtering - if the node is selected, right?

shredlocker: Yep, this is the value we send to the backend if that node is selected ;)

PR #142, opened 2019-02-12
Feature/new selection store

vadymmarkov:
# Why?

In order to make filter nodes independent of selection status.

# What?

- Introduce `CCFilterSelectionStore` class, which keeps selection state, selects / deselects filter nodes and returns additional information based on selection status
- Refactor controllers using `CCFilterSelectionStore`
- Configure cells and views with either view models or actual values instead of passing filter nodes directly

# Show me

No UI changes, hopefully 😅


Sources/CCFilter/ViewControllers/CCRootFilterViewController.swift
```diff
@@ -84,9 +93,31 @@ extension CCRootFilterViewController: UITableViewDelegate {
...
+    func rootFilterCell(_ cell: CCRootFilterCell, didRemoveItemAt index: Int) {
+        guard let indexPath = tableView.indexPath(for: cell) else {
+            return
+        }
+
+        let currentFilterNode = filterNode.children[indexPath.row]
+        let selectedChildren = selectionStore.selectedChildren(for: currentFilterNode)
+
+        selectionStore.deselect(node: selectedChildren[index], withChildren: true)
+        tableView.reloadRows(at: [indexPath], with: .automatic)
```
shredlocker: You don't need to reload the row, it is animated by its own ;)

PR #145, opened 2019-02-13
Show stepper filter for number of bedrooms

shredlocker:
# Why?

Number of bedrooms option was presented as range filter.

# What?

- Added StepperFilterInfo which simply extends RangeFilterInfo to separate between the two filter types.

# Show me

No UI

### DIFF
--- a/Sources/CCFilter/Models/ViewModels/RangeFilterInfo.swift
+++ b/Sources/CCFilter/Models/ViewModels/RangeFilterInfo.swift
@@ -2,7 +2,7 @@
 //  Copyright © FINN.no AS, Inc. All rights reserved.
 //
 
-public final class RangeFilterInfo {
+public class RangeFilterInfo {
     typealias RangeBoundsOffsets = (hasLowerBoundOffset: Bool, hasUpperBoundOffset: Bool)
     typealias AccessibilityValues = (stepIncrement: Int?, valueSuffix: String?)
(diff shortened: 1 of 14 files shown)


Sources/CCFilter/Models/ViewModels/RangeFilterInfo.swift
```diff
@@ -2,7 +2,7 @@
 //  Copyright © FINN.no AS, Inc. All rights reserved.
 //
 
-public final class RangeFilterInfo {
+public class RangeFilterInfo {
     typealias RangeBoundsOffsets = (hasLowerBoundOffset: Bool, hasUpperBoundOffset: Bool)
     typealias AccessibilityValues = (stepIncrement: Int?, valueSuffix: String?)
     typealias AppearenceProperties = (usesSmallNumberInputFont: Bool, displaysUnitInNumberInput: Bool, isCurrencyValueRange: Bool)
```
vadymmarkov: Instead of subclassing I'd better add a property `let kind: Kind`, which is an enum:

```swift
public final class RangeFilterInfo {
   public enum Kind {
      case slider
      case stepper
   }

   init(kind: Kind, ...) { ... }
} 
```

shredlocker: Yes! Good idea 😄

PR #150, opened 2019-02-15
New architecture and selection

shredlocker:
# Why?

This looks terrifying! But it is not really, most of the changes comes from creating new files, copy and paste the content from the old files and delete the old files. This makes it a bit hard to see the changes so I will do my best to describe them.

The complexity of this project has reached a point where it is worth spending some time simplifying the essential parts, i.e. the architecture and selection process. This will make it easier to implement new functionality without increasing the complexity, continue to refactor in the future and to use in the FINN app.

# What?

### Architecture

The current implementation uses navigators to navigate between view controllers. 
```swift
navigator.navigate(to: .multiLevelSelectionListFilter(filterInfo: multiLevelListSelectionFilterInfo))
```
Instead of using navigators we can use the CCFilterViewController to push and pop view controllers. Each view controller has a delegate which will notify of any filter selections. 
```swift
delegate?.viewController(self, didSelect: filterNode)
```
By chaining the delegates together we will achieve an architecture like this:

architecture

Each view controller can react to any filter selections down the chain and when the delegation calls have reached the CCFilterViewController, it can simply push the corresponding view controller based on the selected filter node. This will make the view controller unaware of their position within the filter.
```swift
extension CCFilterViewController: CCViewControllerDelegate {
        func viewController(_ viewController: CCViewController, didSelect filterNode: CCFilterNode) {
                // E.g.
                let listViewController = CCListFilterViewController(filterNode: filterNode)
                listViewController.delegate = viewController
                pushViewController(listViewController, animated: true)
        }
}
```
Create the filter in the app is easier as well. 
```swift
let navigationController = FilterNavigationController()
let factory = FilterDependencyContainer(
        selectionDataSource: selectionDataSource, 
        searchQuerySuggestionsDataSource: nil, 
        filterDelegate: filterDelegate, 
        filterSelectionTitleProvider: FilterSelectionTitleProvider(), 
        mapFilterViewManager: mapFilterViewManager, 
        searchLocationDataSource: CharcoalSearchLocationDataSource(networkClient: networkClient))
let rootFilterNavigator = factory.makeRootFilterNavigator(navigationController: navigationController)
let filterRootStateController = rootFilterNavigator.start()
filterRootStateController .change(to: .loadFreshFilters(data: filter))
return navigationController
```
Now
```swift
let filterViewController = CCFilterViewController(filter: filter, config: config)
filterViewController.filterDelegate = self
filterViewController.mapFilterViewManager = mapFilterViewManager
filterViewController.searchLocationDataSource = CharcoalSearchLocationDataSource(networkClient: networkClient))
filterViewController.searchQuerySuggestionDataSource = nil
return filterViewController
```
CCFilterViewController replaces the dependency container together with the factories. CCFilterViewController is the only view controller exposed to the app.

### Selection

FilterInfo classes and protocols are replaced with CCFilterNodes. The CCFilterViewController is initialised with a CCFilter which contains a reference to the root filter node and the verticals for that filter. We can get a CCFilter from the FilterSetup object by calling its `asCCFilter`method:
```swift
let filter = filterSetup.asCCFilter()
```
Each view controller is then initialised with a specific node within the filter and is responsible for that node and its children. 
Updating the filters can be done by setting the filter property in the CCFilterViewController
```swift
filterViewController.filter = updatedFilter
```
The SelectionDataSource is replaced with CCFilterSelectionStore which is responsible for setting the selection for the filter and getting the selection titles used in the root filter.

### Other Changes

- Created CCListFilterViewController which replaces both list and multi level list. This view controller works nice with the linear architecture.
- Simplified root filter view controller and its views. and made them work with the node structure. We use filter nodes for search and inline filters so these works great with the selection store.


Sources/FINNFilterImplementation/BackendModel/FilterDataQuery.swift
```diff
@@ -28,6 +28,16 @@ struct FilterDataQuery: Decodable {
         return FilterDataQuery(title: title, value: value, totalResults: totalResults ?? 0, filter: filter)
     }
 
+    public func filterNode(name: String) -> CCFilterNode {
```
osanoj: Does it need to be public?

Sources/FINNFilterImplementation/BackendModel/FilterData.swift
```diff
@@ -27,6 +27,20 @@ struct FilterData: Decodable {
         queries = try container.decodeIfPresent([FilterDataQuery].self, forKey: CodingKeys.queries) ?? []
     }
 
+    public func filterNode() -> CCFilterNode {
```
osanoj: Does it need to be public?

Sources/CCFilter/ViewControllers/CCRootFilterViewController.swift
```diff
@@ -0,0 +1,193 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+protocol CCRootFilterViewControllerDelegate: CCViewControllerDelegate {
```
osanoj: Is inheriting another protocol really necessary? Can't you maybe specify that rootDelegate should confirm to both instead?

Sources/CCFilter/ViewControllers/CCFilterViewController.swift
```diff
@@ -0,0 +1,141 @@
...
+            loadingViewController.remove()
+        }
+    }
+}
+
+// MARK: - CCRootFilterViewControllerDelegate
+
+extension CCFilterViewController: CCRootFilterViewControllerDelegate {
+    func rootFilterViewController(_ viewController: CCRootFilterViewController, didSelectVerticalAt index: Int) {
+        filterDelegate?.filterViewController(self, didSelectVerticalAt: index)
```
osanoj: It maybe makes sense to tell the delegate which vertical that was selected, the index is probably a lot less useful?

Sources/CCFilter/ViewControllers/CCFilterViewController.swift
```diff
@@ -0,0 +1,141 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+public protocol CCFilterViewControllerDelegate: class {
```
osanoj: How will the changes in selection and current selection values be propagated to the "owner" of the filter?

PR #154, opened 2019-02-18
Rename models

vadymmarkov:
# Why?

To get better names for models.

# What?

- Rename `CCFilterNode` to `Filter`
- Rename `CCFilter` to `FilterContainer`
- Rename `children` to `subfilters`
- Use `filter` instead of `node` as a variable name


Sources/Core/CharcoalViewController.swift
```diff
@@ -92,36 +92,37 @@ extension CharcoalViewController: FilterViewControllerDelegate {
         popToRootViewController(animated: true)
     }
 
-    func filterViewController(_ viewController: FilterViewController, didSelectFilter filterNode: CCFilterNode) {
-        guard !filterNode.isLeafNode else { return }
+    func filterViewController(_ viewController: FilterViewController, didSelectFilter filter: Filter) {
+        guard !filter.isLeafFilter else { return }
```
shredlocker: Is there a more suitable name for this parameter? I don't feel like `isLeafFilter` works anymore. Maybe it is better to use `filter.subFilters.isEmpty`directly.

vadymmarkov: I agree 😉

vadymmarkov: Fixed now.

PR #140, opened 2019-02-08
Filter selection tags: refactor views

vadymmarkov:
# Why?

Because we had several unnecessary views, protocols and some code duplication in place.

# What?

- Rename `CurrentSelectionValuesContainerView` to `SelectionTagsContainerView`
- Rename `FilterTagView` to `SelectionTagView`
- Remove `CollapsedSelectionValuesView` and `ExpandedSelectionValuesView`
- Simplify implementation

# Show me

No UI changes

PR #155, opened 2019-02-19
Filter selection tags: refactor views

vadymmarkov:
# Why?

Because we had several unnecessary views, protocols and some code duplication in place, same as https://github.com/finn-no/charcoal-ios/pull/140

# What?

- Introduce `SelectionTagView ` and `SelectionTagsContainerView`
- Simplify implementation
- Get back validation for range selections

# Show me

No UI changes


Sources/Core/Selection/Views/SelectionTagsContainerView.swift
```diff
@@ -0,0 +1,107 @@
...
+}
+
+final class SelectionTagsContainerView: UIView {
+    weak var delegate: SelectionTagsContainerViewDelegate?
+
+    // MARK: - Private properties
+
+    private lazy var collapsedView = SelectionTagView(withAutoLayout: true)
+
+    private lazy var expandedView: UIStackView = {
```
shredlocker: I think we could have a better name for this. It is not really expanded, it is just not collapsed :P maybe something like `selectionTagStackView` instead, or simply `stackView`. I feel like this makes it more clear what the `collapsedView` really is.

vadymmarkov: Makes sense. `stackView` it is 👍

PR #156, opened 2019-02-19
Feature/context filters

shredlocker:
# Why?

We need to add support for context filters. This is the very beginning so there is no fancy animations for now.

# What?

- Added Filter.Kind which can either be `.normal` or `.context`
- Added context mark to RootFilterCell
- Configured all context filter keys in BAP and car markets
- Sets the filter kind when converting from FilterSetup to FilterContainer
- Small demo to test FilterSetup convertion

# Show me

simulator screen shot - iphone xs - 2019-02-19 at 13 20 36


Sources/Core/Models/RangeFilter.swift
```diff
@@ -11,10 +11,10 @@ final class RangeFilter: Filter {
     let lowValueFilter: Filter
     let highValueFilter: Filter
 
-    init(title: String, name: String) {
+    init(title: String, name: String, kind: Kind = .normal) {
```
vadymmarkov: Can ranges be context filters?

Sources/Core/Filters/Root/Views/CCRootFilterCell.swift
```diff
@@ -0,0 +1,139 @@
...
+
+    private lazy var titleLabel: UILabel = {
+        let label = UILabel(frame: .zero)
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.font = .regularBody
+        label.textColor = .licorice
+        return label
+    }()
+
+    private lazy var collectionSelectionView: CCCollectionSelectionView = {
```
vadymmarkov: Didn't I remove this?

shredlocker: I've fixed it now. Merging issues ;)

Sources/Core/Models/RangeFilter.swift
shredlocker: Yes, they can. E.g. ski length

Sources/FINN/FilterMarkets/FilterMarketB2B.swift
```diff
@@ -98,6 +98,10 @@ extension FilterMarketB2B: FilterConfiguration {
         }
     }
 
+    public var contextFilters: Set<FilterKey> {
```
vadymmarkov: Not related to this PR, but it's weird that FINN-specific `FilterKey` type is a part of `FilterConfiguration` protocol for general use.

shredlocker: I think this configuration part needs an update at some point.

PR #159, opened 2019-02-19
Refactor filter configuration

vadymmarkov:
# Why?

To make filter configuration more customizable and less FINN-specific.

# What?

- Make it possible to specify keys for the map and range sub filters 
- Changes to filter configuration:
  -  Use `String` instead of `FilterKey` to remove references to FINN-specific types from the "core" files
  - Add `FINNFilterConfiguration` protocol for working with FINN-specific types 
  -  Rename `filterMapKey` to `mapFilterParentFilterKey`
  - Add new `searchFilterKey` and `preferencesFilterKey` properties to use it instead of hard-coded strings
  - Rename `viewModel(forKey:)` to `rangeViewModel(forKey:)`
- `Filter`: rename `name` to `key` to align naming with `FilterKey` enum

# Show me

No UI changes.


vadymmarkov:
I think we could even make two targets in the project, one for Charcoal Core and one for FINN implementation. Then we'll get compiler errors if we use some FINN types in Core.

shredlocker:
I like that!

osanoj:
Sounds like a good idea 👍

PR #161, opened 2019-02-20
Move map filter to the root level and add mutually exclusive filters in configuration

vadymmarkov:
# Why?

In the FINN app "location" and "map" filters are mutually exclusive and it doesn't make sense to select them together. At the same time it's not the best UX to clear out "location" when "map" is selected and the other way around.

# What?

- We decided to move "map" filter to the root level and disable "location" filter  when "map" is selected. It works both ways and not only for those two filters. It's possible to specify your custom mutually exclusive filters in `FilterConfiguration` now.

-  Show radius as a tag to indicate that map filter is selected

- Rename string properties from FilterConfiguration to not confuse the type system

# Show me

simulator screen shot - iphone xs - 2019-02-20 at 15 49 32


Sources/Core/Filters/Root/RootFilterViewController.swift
```diff
@@ -39,7 +39,12 @@ final class RootFilterViewController: FilterViewController {
     }()
 
     private var searchFilter: Filter? {
-        return filter.subfilters.first { $0.key == config.searchFilterKey }
+        return filter.subfilters.first { $0.key == config.searchFilter }
+    }
+
+    private func predicateForMutuallyExclusiveFilters(excluding filter: Filter) -> ((Filter) -> Bool) {
```
shredlocker: This function isn't used anywhere

Sources/FINN/BackendModel/FilterSetup.swift
```diff
@@ -50,33 +50,35 @@ public struct FilterSetup: Decodable {
...
-                locationKey: "geoLocationName"
+                key: key,
+                latitudeKey: FilterKey.latitude.rawValue,
+                longitudeKey: FilterKey.longitude.rawValue,
+                radiusKey: FilterKey.radius.rawValue,
+                locationKey: FilterKey.geoLocationName.rawValue
             )
-            locationFilter.add(subfilter: mapFilter, at: 0)
+
+            let index = supportedFilters.firstIndex(where: { $0.key == config.locationFilter }) ?? 0
```
shredlocker: Could add map key into the `supportedFilters` at the correct index in the config.

Sources/Core/Filters/Root/RootFilterViewController.swift
vadymmarkov: Nice catch, it's a leftover from my experiments 😅

Sources/FINN/BackendModel/FilterSetup.swift
vadymmarkov: Great idea, now I know what's `supportedFilters` for 😄

shredlocker: We could maybe come up with a better name for those filters :P

shredlocker: `RootLevelFilters` or something is a bit more descriptive

vadymmarkov: @shredlocker Should be fixed now 😉

PR #163, opened 2019-02-22
updates protocols

shredlocker:
# Why?

We had some protocol methods that wasn't needed and there were no way to get the query items from the selection store and it seems fit to send these as arguments with the delegate call.

# What?

- Removed `didSelectApply` protocol method. View controllers pop directly to root when the button is pressed. `CharcoalViewController` tells the delegate if the selection store has changed when root filters will be displayed.
- Added query items to `CharcoalViewControllerDelegate` method.
- Added `hasChanges` flag to `FilterSelectionStore`.

# Show me

No UI

### DIFF
--- a/Sources/Core/CharcoalViewController.swift
+++ b/Sources/Core/CharcoalViewController.swift
@@ -84,10 +84,6 @@ public class CharcoalViewController: UINavigationController {
 // MARK: - RootFilterViewControllerDelegate
 
 extension CharcoalViewController: RootFilterViewControllerDelegate {
-    func rootFilterViewControllerDidChangeSelection(_ viewController: RootFilterViewController) {
-        filterDelegate?.charcoalViewController(self, didChangeSelection: selectionStore.queryItems(for: filter.rootFilter))
-    }
-
     func rootFilterViewController(_ viewController: RootFilterViewController, didSelectVerticalAt index: Int) {
         guard let vertical = filter.verticals?[safe: index] else { return }
         filterDelegate?.charcoalViewController(self, didSelect: vertical)
@@ -143,7 +139,13 @@ extension CharcoalViewController: FilterViewControllerDelegate {
 
 extension CharcoalViewController: FilterSelectionStoreDelegate {
     func filterSelectionStoreDidChange(_ selectionStore: FilterSelectionStore) {
-        selectionHasChanged = true
+        if topViewController === rootFilterViewController {
+            // Changes in inline filters or removes selected filters
+            filterDelegate?.charcoalViewController(self, didChangeSelection: selectionStore.queryItems(for: filter.rootFilter))
+            selectionHasChanged = false
+        } else {
+            selectionHasChanged = true
+        }
     }
 }
 
@@ -152,7 +154,7 @@ extension CharcoalViewController: FilterSelectionStoreDelegate {
 extension CharcoalViewController: UINavigationControllerDelegate {
     public func navigationController(_ navigationController: UINavigationController, willShow viewController: UIViewController, animated: Bool) {
         guard viewController === rootFilterViewController else { return }
-
+        // Will return to root filters
         if selectionHasChanged {
             filterDelegate?.charcoalViewController(self, didChangeSelection: selectionStore.queryItems(for: filter.rootFilter))
             selectionHasChanged = false
--- a/Sources/Core/CharcoalViewController.swift
+++ b/Sources/Core/CharcoalViewController.swift
@@ -34,6 +34,7 @@ public class CharcoalViewController: UINavigationController {
 
     // MARK: - Private properties
 
+    private var selectionHasChanged = false
     private let selectionStore: FilterSelectionStore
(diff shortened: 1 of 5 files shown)


Sources/Core/Filters/FilterViewController.swift
```diff
@@ -83,6 +82,6 @@ class FilterViewController: UIViewController, FilterBottomButtonViewDelegate {
     // MARK: - FilterBottomButtonViewDelegate
 
     func filterBottomButtonView(_ filterBottomButtonView: FilterBottomButtonView, didTapButton button: UIButton) {
-        delegate?.filterViewControllerDidSelectApply(self)
+        navigationController?.popToRootViewController(animated: true)
```
vadymmarkov: I don't think it's a good practice to access `navigationController` from view controller because then we assume that it was pushed into navigation stack. Nothing will happen if we call `popToRootViewController` from a presented view controller for instance. That's why I'd better call this delegate method in order to keep view controller unaware of navigation.

Sources/Core/Selection/FilterSelectionStore.swift
```diff
@@ -5,6 +5,13 @@
 import Foundation
 
 final class FilterSelectionStore {
+
+    // MARK: - Internal properties
+
+    internal var hasChanges = false
```
vadymmarkov: Don't really like this flag here. How about adding a delegate or observer to `FilterSelectionStore`? Then in `CharcoalViewController` you can subscribe to changes and call appropriate `CharcoalViewControllerDelegate` method. There still could be a boolean flag in `CharcoalViewController` to call its delegate only when the user comes back to the root and so on.

Sources/Core/Filters/FilterViewController.swift
shredlocker: 👍

Sources/Core/Selection/FilterSelectionStore.swift
shredlocker: Sounds good :)

shredlocker: It might be better now :P

Sources/Core/CharcoalViewController.swift
```diff
@@ -136,9 +141,24 @@ extension CharcoalViewController: FilterViewControllerDelegate {
     }
 }
 
+extension CharcoalViewController: FilterSelectionStoreDelegate {
+    func filterSelectionStoreDidChange(_ selectionStore: FilterSelectionStore) {
+        selectionHasChanged = true
```
vadymmarkov: How about:
```swift
if topViewController === rootFilterViewController {
    selectionHasChanged = false
    filterDelegate?.charcoalViewController(self, didChangeSelection: selectionStore.queryItems(for: filter.rootFilter))
} else {
    selectionHasChanged = true
}
```

The you don't need `rootFilterViewControllerDidChangeSelection` anymore and rely only on observing changes in `FilterSelectionStore`.

shredlocker: Good idea! :D

Sources/Core/Selection/FilterSelectionStore.swift
```diff
@@ -66,6 +82,8 @@ extension FilterSelectionStore {
         } else {
             setValue(filter.value, for: filter)
         }
+
+        delegate?.filterSelectionStoreDidChange(self)
```
vadymmarkov: Is it ok that this delegate method will be called at least twice: once in either `removeValues` or `setValue` and then here in `toggleValue`? Also when we remove value for filter it will iterate through sub-filters and make delegate calls for each of them.

shredlocker: I think we should try to avoid this. Maybe we could make the recursive methods private?

Sources/Core/Selection/FilterSelectionStore.swift
```diff
@@ -81,6 +86,27 @@ extension FilterSelectionStore {
     }
 }
 
+private extension FilterSelectionStore {
+    func _setValue<T: LosslessStringConvertible>(_ value: T?, for filter: Filter) {
+        removeValues(for: filter)
```
vadymmarkov: Shouldn't you call `_removeValues` here?

Sources/Core/Selection/FilterSelectionStore.swift
```diff
@@ -81,6 +86,27 @@ extension FilterSelectionStore {
...
+        }
+    }
+
+    func _removeValues(for filter: Filter) {
+        if let queryItem = queryItem(for: filter) {
+            queryItems.remove(queryItem)
+        }
+
+        filter.subfilters.forEach {
+            removeValues(for: $0)
```
vadymmarkov: Shouldn't it be `_removeValues`?

Sources/Core/Selection/FilterSelectionStore.swift
shredlocker: Yes! It should :P

PR #165, opened 2019-02-26
Refactor free text filter

shredlocker:
# Why?

Had to rename from search query filter to free text filter.
`SearchQuerySuggestionDataSource` was a bit weird.

# What?

- Renamed all `SearchQuery` filters to `FreeText`
- Replaced `SearchQuerySuggestionDataSource` with a data source more related to `UITableViewDataSource` and a delegate to notify when the search bar text has changed in order to fetch new suggestions and reload the table view.
- Added filter property to `FreeTextFilterViewController` and added the search filter as an argument to the delegate call. This way we don't have to find the search filter in root every time we need it. (I think we should do the same with the inline filter)
- Added `FreeTextFilterDataSource` and `FreeTextFilterDelegate` to demo

# Show me

 No UI

### DIFF
--- a/Sources/Core/Filters/FreeText/FreeTextFilterViewController.swift
+++ b/Sources/Core/Filters/FreeText/FreeTextFilterViewController.swift
@@ -5,26 +5,35 @@
 import FinniversKit
 import UIKit
 
-public protocol SearchQuerySuggestionsDataSource: AnyObject {
-    func searchQueryViewController(_ searchQueryViewController: SearchQueryViewController, didRequestSuggestionsFor searchQuery: String, completion: @escaping ((_ text: String, _ suggestions: [String]) -> Void))
+public protocol FreeTextFilterDataSource: class {
+    func freeTextFilterTableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int
+    func freeTextFilterTableView(_ tableView: UITableView, titleForCellAt indexPath: IndexPath) -> String
 }
 
-public protocol SearchViewControllerDelegate: class {
-    func presentSearchViewController(_ searchViewController: SearchQueryViewController)
-    func searchViewControllerDidCancelSearch(_ searchViewController: SearchQueryViewController)
-    func searchViewController(_ searchViewController: SearchQueryViewController, didSelectQuery query: String)
+public protocol FreeTextFilterDelegate: class {
+    func freeTextFilterTableView(_ tableView: UITableView, didChangeText text: String?)
 }
 
-public class SearchQueryViewController: UIViewController {
+// Internal protocol to delegate back to root filter view controller
+protocol FreeTextFilterViewControllerDelegate: class {
+    func freeTextFilterViewControllerWillBeginEditing(_ viewController: FreeTextFilterViewController)
+    func freeTextFilterViewControllerWillEndEditing(_ viewController: FreeTextFilterViewController)
+    func freeTextFilterViewController(_ viewController: FreeTextFilterViewController, didSelectValue value: String?, forFilter filter: Filter)
+}
+
+class FreeTextFilterViewController: UIViewController {
 
     // MARK: - Public Properties
 
-    public weak var delegate: SearchViewControllerDelegate?
-    public var searchQuerySuggestionDataSource: SearchQuerySuggestionsDataSource?
+    weak var filterDelegate: FreeTextFilterDelegate?
+    weak var filterDataSource: FreeTextFilterDataSource?
+
+    weak var delegate: FreeTextFilterViewControllerDelegate?
+
+    var filter: Filter?
 
     // MARK: - Private Properties
 
-    private var suggestions: [String] = []
     private var currentQuery: String?
     private var didClearText = false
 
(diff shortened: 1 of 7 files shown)


Sources/Core/Filters/FreeText/FreeTextFilterViewController.swift
```diff
@@ -5,26 +5,35 @@
 import FinniversKit
 import UIKit
 
-public protocol SearchQuerySuggestionsDataSource: AnyObject {
-    func searchQueryViewController(_ searchQueryViewController: SearchQueryViewController, didRequestSuggestionsFor searchQuery: String, completion: @escaping ((_ text: String, _ suggestions: [String]) -> Void))
+public protocol FreeTextFilterDataSource: class {
+    func freeTextFilterTableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int
```
osanoj: Sending the tableView with the delegate exposes unnecessary implementation detail to the delegate. Delegate should not need/care about what kind of view is used to show it? It also makes it difficult to change to another view type to show suggestions in.

shredlocker: I tried to keep the FreeTextViewController internal, but I guess it makes sense to make it public when it is going to call the main app anyways. I'll make it a bit more custom ;)

osanoj: You could make a small protocol that is public maybe, all you need is a method to tell that suggestions are ready, right? Or go back to the completion handler that was there. 

I think you might need to have the connection between search text and suggestions anyway... You might have timing issues otherwise, i.e. suggestions for "iPh" might take longer to come back than suggestions for "iPho". Or queuing it and cancelling old request.

osanoj: But actually, rather remove all the suggestion things for now then spend a lot of time on it. It is not needed for initial release.

shredlocker: Allright :)

Sources/Core/Filters/FreeText/FreeTextFilterViewController.swift
```diff
@@ -5,26 +5,35 @@
 import FinniversKit
 import UIKit
 
-public protocol SearchQuerySuggestionsDataSource: AnyObject {
-    func searchQueryViewController(_ searchQueryViewController: SearchQueryViewController, didRequestSuggestionsFor searchQuery: String, completion: @escaping ((_ text: String, _ suggestions: [String]) -> Void))
+public protocol FreeTextFilterDataSource: class {
+    func numberOfSuggestions(in freeTextFilterViewController: FreeTextFilterViewController) -> Int
+    func freeTextFilterViewController(_ freeTextFilterViewController: FreeTextFilterViewController, suggestionForCellAt indexPath: IndexPath) -> String
```
vadymmarkov: Maybe `suggestionAt` instead of `suggestionForCellAt`?

Sources/Core/Filters/Root/RootFilterViewController.swift
```diff
@@ -190,27 +180,22 @@ extension RootFilterViewController: VerticalListViewControllerDelegate {
...
+        viewController.remove()
         tableView.reloadData()
     }
 
-    func searchViewController(_ searchViewController: SearchQueryViewController, didSelectQuery query: String) {
-        guard let searchFilter = searchFilter else {
-            return
+    func freeTextFilterViewController(_ viewController: FreeTextFilterViewController, didSelectValue value: String?, forFilter filter: Filter) {
+        if let value = value {
+            selectionStore.setValue(value, for: filter)
```
vadymmarkov: You can just `selectionStore.setValue(value, for: filter)`, it will be removed if nil.

vadymmarkov: Also, maybe `FreeTextFilterViewController` can be created with selection store and set it itself. That's what we do for range, map and list at least.

Sources/Core/Filters/Root/RootFilterViewController.swift
```diff
@@ -98,9 +88,9 @@ extension RootFilterViewController: UITableViewDataSource {
 
         switch currentFilter.kind {
         case .search:
-            let cell = tableView.dequeue(SearchQueryCell.self, for: indexPath)
-            cell.searchBar = searchQueryViewController.searchBar
-            cell.searchBar?.placeholder = currentFilter.title
+            let cell = tableView.dequeue(FreeTextFilterCell.self, for: indexPath)
+            cell.configure(with: freeTextFilterViewController.searchBar)
+            freeTextFilterViewController.setup(with: currentFilter)
```
vadymmarkov: It's nice we got rid of `searchFilter` property, but maybe we can create `FreeTextFilterViewController` with filter here? Then we don't have to unwrap optional `filter` in `FreeTextFilterViewController`.

Sources/Core/Filters/Root/RootFilterViewController.swift
shredlocker: Yes, we could do that

Sources/Core/Filters/Root/RootFilterViewController.swift
shredlocker: I can do that :)

PR #168, opened 2019-03-01
Feature/optional filter

shredlocker:
# Why?

We might want to present the `CharcoalViewController` without a filter, but with a loading indicator.

# What?

- Made filter and config private optionals
- Added public method to configure `CharcoalViewController` with filter, config and query items.

### DIFF
--- a/Sources/Core/CharcoalViewController.swift
+++ b/Sources/Core/CharcoalViewController.swift
@@ -14,26 +14,18 @@ public class CharcoalViewController: UINavigationController {
 
     // MARK: - Public properties
 
-    public var filter: FilterContainer {
-        didSet {
-            rootFilterViewController.set(filter: filter.rootFilter, verticals: filter.verticals)
-            isLoading = false
-        }
-    }
-
-    public var config: FilterConfiguration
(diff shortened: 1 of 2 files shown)


Sources/Core/CharcoalViewController.swift
```diff
@@ -49,28 +41,20 @@ public class CharcoalViewController: UINavigationController {
...
-    private let selectionStore: FilterSelectionStore
-    private var rootFilterViewController: RootFilterViewController
+    private var selectionStore = FilterSelectionStore()
 
+    private var rootFilterViewController: RootFilterViewController?
     private lazy var loadingViewController = LoadingViewController(backgroundColor: .milk, presentationDelay: 0)
 
     // MARK: - Init
 
-    public init(filter: FilterContainer, config: FilterConfiguration, queryItems: Set<URLQueryItem> = []) {
```
vadymmarkov: Could we keep `config` in `init`? Isn't it created only once?

shredlocker: When we change submarket we change both the filter and config. We could potentially make one config for all markets and submarkets.

PR #172, opened 2019-03-06
Refactor range info

vadymmarkov:
# Why?

To make it easier to configure range sliders and steppers.

# What?

- Rename `RangeFilterInfo` to `RangeFilterConfiguration` and combined it with `StepSliderInfo`
- Introduce `StepperFilterConfiguration` as a separate configuration type for stepper
- Update filter configuration to use new types
- Use `RangeFilterConfiguration` and `StepperFilterConfiguration` as a part of filter kind. It makes it more predictable when it comes to creation of range and stepper view controllers, because we don't display filters not configured to be ranges or steppers.

# Show me

No UI changes


shredlocker:
Is there a way to handle the mutually exclusive filter without using the `FilterConfiguration`? I seems like that is the only place we use the config after these changes. It would be nice to remove the config from `CharcoalViewController` and `RootFilterViewController`

vadymmarkov:
We can potentially create `RootFilterViewController` with `mutuallyExclusiveFilters: Set<String>` instead of `config`. In `CharcoalViewController we can still keep `config` in init method, but store only `mutuallyExclusiveFilters` to local var. What do you think?

shredlocker:
Sounds good for now ;)

shredlocker:
Or, add `root` in `Filter.Kind` with the mutually exclusive filters

shredlocker:
I think this looks good! Good job 😄

PR #173, opened 2019-03-07
Merging two filters

shredlocker:
# Why?

We need to be able to merge two filters together because some search responses doesn't contain the entire filter.

# What?

- Added `merge(with other: Filter)` function which merges the sub filters of each filter together preserving the index of each sub filter.
- Made `Filter` extend `Equatable` protocol to compare two filter.
- It is now possible to reset the number of hits in each filter. 
- Some tests

# Show me

No UI

### DIFF
--- a/Sources/Core/Models/Filter.swift
+++ b/Sources/Core/Models/Filter.swift
@@ -27,7 +27,7 @@ final class Filter {
     let style: Style
     let kind: Kind
 
-    var subfilters: [Filter] = []
+    private(set) var subfilters: [Filter] = []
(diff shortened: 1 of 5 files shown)


Sources/Core/Models/Filter.swift
```diff
@@ -44,19 +43,40 @@ final class Filter {
 
     // MARK: - Public methods
 
-    private func add(subfilter: Filter, at index: Int? = nil) {
+    func add(subfilter: Filter, at index: Int? = nil) {
```
vadymmarkov: Why can't it be private?

PR #174, opened 2019-03-07
Add support for dynamic price range

vadymmarkov:
# Why?

Because we want to configure some ranges with dynamic step increment.

# What?

- Add new `ValueKind.intervals(array: [StepInterval], defaultIncrement: Int)`, which can be used like this:

```swift
.intervals(
    array: [
        (range: 0 ..< 500, increment: 50), // 0, 50, 100, 150, ..., 450
        (range: 500 ..< 1500, increment: 100), // 500, 600, 700, 800, ..., 1400
        (range: 1500 ..< 6000, increment: 500), // 1500, 2000, 2500, ..., 5500 
    ],
    defaultIncrement: 1000 // 6000, 7000, 8000
)
```

- Configure dynamic price range for Torget according to requirements 

# Show me

No UI changes.

### DIFF
--- a/Sources/Core/Models/RangeFilterConfiguration.swift
+++ b/Sources/Core/Models/RangeFilterConfiguration.swift
@@ -3,9 +3,12 @@
 //
 
 public struct RangeFilterConfiguration: Equatable {
+    public typealias StepInterval = (range: Range<Int>, increment: Int)
+
     public enum ValueKind {
         case incremented(Int)
         case steps([Int])
+        case intervals(array: [StepInterval], defaultIncrement: Int)
     }
 
     public let minimumValue: Int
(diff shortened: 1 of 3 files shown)


Sources/Core/Models/RangeFilterConfiguration.swift
```diff
@@ -3,9 +3,12 @@
 //
 
 public struct RangeFilterConfiguration: Equatable {
+    public typealias StepInterval = (range: Range<Int>, increment: Int)
+
     public enum ValueKind {
         case incremented(Int)
         case steps([Int])
+        case intervals(array: [StepInterval], defaultIncrement: Int)
```
shredlocker: I think I would prefer to use an interval with an open range instead of using default increment. Like this: 
```swift
[
    (range: 0 ..< 500, increment: 50),
    (range: 500 ... , increment: 100)
]
```
Do you think there are situations where the intervals are not continues?

vadymmarkov: I agree it looks better, but `500 ...` is  `PartialRangeFrom` type which is not convertible to `Range`. We could of course introduce a new type instead:

```swift
struct StepInterval {
   let from: Int
   let to: Int
   let increment: Int
}
```

shredlocker: Right! I'm not sure which I like better :P It is nice to use the `Range` operator.. Could perform some Swift magic and make our own operator, or overload the `...` :P

shredlocker: What about using an array of `(from: Int, increment: Int)` and define a range between the elements of the array? if it is always going to be continues it will be fine. I don't know if it is better though...

vadymmarkov: Great idea! I think it's easier to read 👍

vadymmarkov:
@shredlocker Fixed according to your last comment 😉

PR #176, opened 2019-03-08
Refactor selection for list filters

vadymmarkov:
# Why?

Because we want to change the logic behind "Alle" button and improve UX when it comes to selection. 

# What?

- Disable/enable subfilters when the user selects/deselects "Alle" button
- Add custom selection animation for cell with animated checkboxes (default `deselectedRows` method from `UItableView` doesn't play well with custom checkbox animations we have)
- Update `FinniversKit`
- Don't select filter through subfilters if it's not a map or range filter
- Fix bottom cells being covered by "Bruk" button when selected/deselected

# Show me

selectall


Sources/Core/Filters/List/Views/ListFilterCell.swift
```diff
@@ -0,0 +1,145 @@
...
+            isUserInteractionEnabled = isEnabled
+            checkboxImageView.isEnabled = isEnabled
+            overlayView.isHidden = isEnabled
+            bringSubviewToFront(overlayView)
+        }
+    }
+
+    private lazy var checkboxImageView = ListFilterImageView(withAutoLayout: true)
+
+    private lazy var chevronImageView: UIImageView = {
```
shredlocker: I was a bit confused about this name. Maybe `alternativeAccessoryIconView`, or something a bit shorter :P

vadymmarkov: I've managed to use default `accessoryView` from table view cell. Now it should even less confusing.

shredlocker:
Nice!

PR #177, opened 2019-03-08
Seperate filter selection configuration

shredlocker:
# Why?

It was a bit weird to use the same function to set both filter and query items. Separating these makes more sense. And I added delegate method for show result button pressed.

# What?

- Removed query items from `configure(with filter: Filter, queryItems: Set<URLQueryItem>)` function.
- Added method to set a selection.
- Added delegate method for when the show results button is pressed. 

# Show me

No UI

### DIFF
--- a/Sources/Core/CharcoalViewController.swift
+++ b/Sources/Core/CharcoalViewController.swift
@@ -73,11 +73,8 @@ public class CharcoalViewController: UINavigationController {
     // MARK: - Public
 
     public func set(selection: Set<URLQueryItem>) {
-        selectionStore = FilterSelectionStore(queryItems: selection)
-
-        if let filter = filter {
-            rootFilterViewController?.set(filter: filter.rootFilter, verticals: filter.verticals)
-        }
+        selectionStore.set(selection: selection)
+        rootFilterViewController?.reloadFilters()
     }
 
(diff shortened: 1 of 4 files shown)


Sources/Core/CharcoalViewController.swift
```diff
@@ -67,6 +70,16 @@ public class CharcoalViewController: UINavigationController {
         delegate = self
     }
 
+    // MARK: - Public
+
+    public func set(selection: Set<URLQueryItem>) {
+        selectionStore = FilterSelectionStore(queryItems: selection)
```
vadymmarkov: If you create a new instance of selection store here it needs to be injected to `rootFilterViewController` and potentially more view controllers. Even though I don't really like it, I think it would be easier to create selection store only once and them set query items by making `var queryItems` internal.

shredlocker: Je, that's true. We could have a `set(selection: ...)` in FilterSelectionStore as well and keep `queryItems` private.

vadymmarkov: Yep, that's even better 👍

shredlocker: Fixed now ;)

PR #179, opened 2019-03-08
Fix inline filter selection

shredlocker:
# Why?

The inline filter selection wasn't working properly.

# What?

- Added `filter`, `vertical` and `selectionStore` properties to `InlineFilterView`.
- Segments will be setup with selected states from `selectionStore`
- Selection is done in `InlineFilterView` directly.
- Removed `didChangeSegment` delegate method
- Refactored `InlineFilterCell` in `RootFilterViewController`

# Show me

No UI


Sources/Core/Filters/Inline/InlineFilterView.swift
```diff
@@ -97,11 +107,25 @@ private extension InlineFilterView {
             segments.append(segment)
         }
 
-        for titles in segmentTitles {
-            let segment = Segment(titles: titles)
-            segment.addTarget(self, action: #selector(handleValueChanged(segment:)), for: .valueChanged)
-            segments.append(segment)
+        if let filter = filter {
```
vadymmarkov: Can we keep `InlineFilterView` unaware of selection store, filter, vertical and other model-related types? If the lines below are the ones that fix inline filter selection, they could perhaps go to `RootFilterViewController`.

shredlocker: Sure, fixing it ;)

shredlocker: Changed it back now. I also moved the `didChangeSelection` delegate call into `didSelectFilter` because the inline filter would remove and set multiple filter, which resulted in multiple search requests. Now it will only be one request. Had to do some changes in `FreeTextFilterViewController` to use the same method.

Demo/Components/InlineFilterDemoViewController.swift
```diff
@@ -6,7 +6,7 @@
 
 final class InlineFilterDemoViewController: UIViewController {
     lazy var inlineFilterView: InlineFilterView = {
-        let view = InlineFilterView()
+        let view = InlineFilterView(frame: .zero)
```
vadymmarkov: Can we use `InlineFilterView(withAutoLayout: true)`?

Sources/Core/Filters/Inline/InlineFilterView.swift
```diff
@@ -5,26 +5,20 @@
...
-        didSet {
-            setupItems()
-        }
-    }
-
     weak var delegate: InlineFilterViewDelegate?
 
     // MARK: - Private properties
 
+    private var vertical: String?
```
vadymmarkov: I think we can remove those two properties and merge `setupItems` with `configure` method where we have all the data needed to build segments.

Sources/Core/Filters/Root/RootFilterViewController.swift
```diff
@@ -209,15 +220,25 @@ extension RootFilterViewController: RootFilterCellDelegate {
 // MARK: - CCInlineFilterViewDelegate
 
 extension RootFilterViewController: InlineFilterViewDelegate {
-    func inlineFilterView(_ inlineFilterView: InlineFilterView, didChangeSegment segment: Segment, at index: Int) {
-        guard let subfilter = filter.subfilter(at: index) else { return }
+    func inlineFilterView(_ inlineFilteView: InlineFilterView, didChange segment: Segment, at index: Int) {
+        guard let inlineFilter = filter.subfilters.first(where: { filter -> Bool in
+            if case .inline = filter.kind {
```
vadymmarkov: We could make kind `Equatable` to make it easier to check for it.

Demo/Components/InlineFilterDemoViewController.swift
shredlocker: Yep ;)

Sources/Core/Filters/Inline/InlineFilterView.swift
shredlocker: Good idea! :)

PR #182, opened 2019-03-12
Implement event logging

vadymmarkov:
# Why?

Because we want to log different UI events.

# What?

- Add `EventLogging` protocol that can be implemented to receive UI events from `Charcoal`:

```swift
final class EventLogger: EventLogging {
    func log(event: Event) {
        print("Event occured: \(event)")
    }
}

charcoalViewController.eventLogger = EventLogger()
```

- Add events related to filter selection and range sliders. More events will be added in future when needed.

# Show me

No UI changes.


vadymmarkov:
We decided to not log UI events for now, but I'm going to use some of the code for different purposes.

PR #183, opened 2019-03-12
Fix inline filter bugs

shredlocker:
# Why?

- Selecting filters was wrong when there were both vertical and filter segments. The index was wrong.
- Auto sizing cells caused inline filter to shift when reloading data. 

# What?

- Added `Section` enum to `InlineFilterView` to separate vertical and filter segments. Adjusted `UICollectionViewDelegateFlowLayout` to make the UI correct.
- Removed `estimatedItemSize` and replaced with `sizeForItem` delegation call to set item size.
- Made `SegmentButton` an internal class

# Show me

No UI changes

### DIFF
--- a/Sources/Charcoal/Filters/Inline/SegmentButton.swift
+++ b/Sources/Charcoal/Filters/Inline/SegmentButton.swift
@@ -12,7 +12,7 @@ extension SegmentButton {
 
 class SegmentButton: UIButton {
     static let borderColor: UIColor = .silver
-    static let borderWidth = 1.5 as CGFloat
+    static let borderWidth: CGFloat = 1.5
 
     var borderStyle: BorderStyle = .middle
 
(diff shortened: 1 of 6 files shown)


Sources/Charcoal/Filters/Inline/SegmentButton.swift
```diff
@@ -10,19 +10,19 @@ extension SegmentButton {
     }
 }
 
-public class SegmentButton: UIButton {
-    public static let borderColor: UIColor = .silver
-    public static let borderWidth = 1.5 as CGFloat
+class SegmentButton: UIButton {
+    static let borderColor: UIColor = .silver
+    static let borderWidth = 1.5 as CGFloat
```
vadymmarkov: `borderWidth: CGFloat  = 1.5` is better.

shredlocker: Consider it fixed 😉

PR #186, opened 2019-03-13
Notify filterDelegate when text field is editing

shredlocker:
# Why?

We need a way to tell the FINN app when the user edits text because the bottom sheet should then be expanded.

# What?

- Added `WillBeginTextEditing` and `WillEndTextEditing` methods to `FilterViewControllerDelegate` and `CharcoalViewControllerDelegate`.
- Added bottom sheet logic in the demo project
- Implemented this in `MapFilterViewController`. We are not going to use it in root for now.

# Show me

test

### DIFF
--- a/Sources/Charcoal/CharcoalViewController.swift
+++ b/Sources/Charcoal/CharcoalViewController.swift
@@ -4,12 +4,15 @@
 
 import UIKit
 
-public protocol CharcoalViewControllerDelegate: class {
+public protocol CharcoalViewControllerTextEditingDelegate: AnyObject {
+    func charcoalViewControllerWillBeginTextEditing(_ viewController: CharcoalViewController)
+    func charcoalViewControllerWillEndTextEditing(_ viewController: CharcoalViewController)
+}
+
+public protocol CharcoalViewControllerSelectionDelegate: AnyObject {
     func charcoalViewController(_ viewController: CharcoalViewController, didSelect vertical: Vertical)
     func charcoalViewController(_ viewController: CharcoalViewController, didSelectExternalFilterWithKey key: String, value: String?)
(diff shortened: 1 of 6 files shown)


Sources/Charcoal/CharcoalViewController.swift
```diff
@@ -8,6 +8,8 @@ public protocol CharcoalViewControllerDelegate: class {
     func charcoalViewController(_ viewController: CharcoalViewController, didSelect vertical: Vertical)
     func charcoalViewController(_ viewController: CharcoalViewController, didSelectExternalFilterWithKey key: String, value: String?)
     func charcoalViewControllerDidPressShowResults(_ viewController: CharcoalViewController)
+    func charcoalViewControllerWillBeginTextEditing(_ viewController: CharcoalViewController)
+    func charcoalViewControllerWillEndTextEditing(_ viewController: CharcoalViewController)
     func charcoalViewController(_ viewController: CharcoalViewController,
```
vadymmarkov: Maybe we can split this protocol to `CharcoalViewControllerSelectionDelegate` and `CharcoalViewControllerTextEditingDelegate`?

shredlocker: Yes, we can do that. The same with `FilterViewControllerDelegate`? Have a separate protocol like `TextEditingDelegate` or something

vadymmarkov: I think we can keep `FilterViewControllerDelegate` as long as it works for as because it's an internal protocol. But `CharcoalViewControllerDelegate` on the other hand is public, so it might be more flexible if we split it in two.

shredlocker: Fixed

PR #195, opened 2019-03-20
Map filter:  change the logic for finding user location

vadymmarkov:
# Why?

Because there were a few bugs in the map filter.

# What?

- Change the logic for finding user location
- Always show location button, even when location services are disabled
- Remove pulsating animations
- Show area name instead of radius in selection tag
- Don't center on user location on enter 

# Show me

No UI changes.

### DIFF
--- a/Sources/Charcoal/Filters/Map/MapView/MapFilterView.swift
+++ b/Sources/Charcoal/Filters/Map/MapView/MapFilterView.swift
@@ -41,9 +41,9 @@ final class MapFilterView: UIView {
         return mapView.centerCoordinate
     }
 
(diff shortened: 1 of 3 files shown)


Sources/Charcoal/Filters/Map/MapView/MapFilterView.swift
```diff
@@ -41,16 +41,14 @@ final class MapFilterView: UIView {
         return mapView.centerCoordinate
     }
 
-    var isUserLocatonButtonEnabled = false {
+    var isUserLocatonButtonHighlighted = false {
```
shredlocker: Missing the `i` in `Locaton`

PR #201, opened 2019-03-26
Add unit and + to stepper filter selection title

shredlocker:
# Why?

Stepper filter selection title is only `3`, we want it to be `3+ soverom`

# What?

- Added alternative unit to stepper configuration.
- Use the alternative unit to display current stepper filter selection.

# Show me

### Before

Screenshot 2019-03-26 at 14 09 44

### After

Screenshot 2019-03-26 at 14 07 54

### DIFF
--- a/Sources/Charcoal/Selection/FilterSelectionStore.swift
+++ b/Sources/Charcoal/Selection/FilterSelectionStore.swift
@@ -140,14 +140,14 @@ extension FilterSelectionStore {
                 return ["\(lowValue ?? "...") - \(highValue ?? "...")"]
             }
         case .stepper:
-            guard let lowValue: Int = value(for: filter) else { return [] }
-            return ["\(lowValue)+"]
-        case let .map(_, _, radiusFilter, locationNameFilter):
-            let locationName: String? = value(for: locationNameFilter)
-            let radius: Int? = value(for: radiusFilter)
-
-            if let title = locationName ?? radius.map({ MapDistanceValueFormatter().title(for: $0) }) {
-                return [title]
+            if let lowValue: Int = value(for: filter) {
+              return ["\(lowValue)+"]
+            } else {
+              return []
+            }
+        case let .map(_, _, radiusFilter, _):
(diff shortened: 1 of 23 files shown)


Sources/FINNSetup/FilterMarkets/FilterMarketRealestate.swift
```diff
@@ -270,7 +270,7 @@ extension FilterMarketRealestate: FINNFilterConfiguration {
 
         switch filterKey {
         case .noOfBedrooms:
-            return StepperFilterConfiguration(minimumValue: 0, maximumValue: 6, unit: "stk")
+            return StepperFilterConfiguration(minimumValue: 0, maximumValue: 6, unit: "stk", alternativeUnit: "soverom")
```
vadymmarkov: Why don't we use "stk" for bedrooms?

shredlocker: I asked on Trello, I hope they say we can use "stk".

shredlocker: Now it only has the `+` at the end.

Sources/Charcoal/Selection/FilterSelectionStore.swift
```diff
@@ -140,7 +140,8 @@ extension FilterSelectionStore {
                 return ["\(lowValue ?? "...") - \(highValue ?? "...")"]
             }
         case .stepper:
-            return value(for: filter).map({ [$0] }) ?? []
+            guard let lowValue: Int = value(for: filter) else { return [] }
+            return ["\(lowValue)+"]
```
vadymmarkov: `return value(for: filter).map({ ["\($0)+"] }) ?? []` should also work, but keep `guard` if you like it more 😉

shredlocker: I changed if `if let` instead to be more consistent with the other cases. That map thing is a nice one liner though 😄

PR #202, opened 2019-03-26
Fix bottom button appearance

vadymmarkov:
# Why?

Because we want to show the bottom button across the view controllers in navigation stack if there is a change in filter selection.

# What?

- Show bottom button across the view controllers in navigation stack
- Remove separator on the last cell in list and root filters
- Enable/disable list filter cells through view model
- Add `UITests` in order to test bottom button and future UI fixes or changes

# Show me

filters


vadymmarkov:
Yep, now we should add more of them 😄

PR #207, opened 2019-03-27
Fix scroll to top on reset

shredlocker:
# Why?

We should only scroll to top when tapped the reset button. Inline filter should also scroll to left.

# What?

- Removed public function `returnToRoot`.
- Added properties `inlineFilterCell` and `freeTextFilterCell` since these are not reused and it is alot easier to scroll the inline filter view with this property.
- Scrolls both root view and inline filter view when reset button is tapped.

# Show me

No UI

### DIFF
--- a/Sources/Charcoal/Filters/Root/RootFilterViewController.swift
+++ b/Sources/Charcoal/Filters/Root/RootFilterViewController.swift
@@ -44,7 +44,20 @@ final class RootFilterViewController: FilterViewController {
         return button
     }()
 
-    private var freeTextFilterViewController: FreeTextFilterViewController?
+    private lazy var freeTextFilterCell: FreeTextFilterCell = {
+        guard let freeTextFilter = filter.subfilters.first(where: { $0.kind == .search }) else { fatalError("Missing free text filter") }
+        let freeTextViewController = FreeTextFilterViewController(filter: freeTextFilter, selectionStore: selectionStore)
+        freeTextViewController.delegate = self
+        let cell = FreeTextFilterCell(viewController: freeTextViewController)
+        return cell
+    }()
+
+    private lazy var inlineFilterCell: InlineFilterCell = {
+        let inlineFilterView = InlineFilterView(frame: .zero)
+        inlineFilterView.delegate = self
+        let cell = InlineFilterCell(view: inlineFilterView)
+        return cell
+    }()
 
     // MARK: - Filter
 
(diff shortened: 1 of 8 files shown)


Sources/Charcoal/Filters/Root/RootFilterViewController.swift
```diff
@@ -44,7 +44,20 @@ final class RootFilterViewController: FilterViewController {
...
-    private var freeTextFilterViewController: FreeTextFilterViewController?
+    private lazy var freeTextFilterCell: FreeTextFilterCell = {
+        guard let freeTextFilter = filter.subfilters.first(where: { $0.kind == .search }) else { fatalError("Missing free text filter") }
+        let freeTextViewController = FreeTextFilterViewController(filter: freeTextFilter, selectionStore: selectionStore)
+        freeTextViewController.delegate = self
+        let cell = FreeTextFilterCell(viewController: freeTextViewController)
+        return cell
+    }()
+
+    private lazy var inlineFilterCell: InlineFilterCell = {
```
vadymmarkov: I'd keep both `FreeTextFilterCell` and `InlineFilterCell` as reusable cells because out API allows having multiple inline filters, even though we don't do that in our filter setup. If you want to access `InlineFilterCell` you can for example iterate through visible cells, cast them to `InlineFilterCell` and call `resetContentOffset`. What do you think?

Sources/Charcoal/Filters/Root/RootFilterViewController.swift
```diff
@@ -44,7 +44,20 @@ final class RootFilterViewController: FilterViewController {
         return button
     }()
 
-    private var freeTextFilterViewController: FreeTextFilterViewController?
+    private lazy var freeTextFilterCell: FreeTextFilterCell = {
+        guard let freeTextFilter = filter.subfilters.first(where: { $0.kind == .search }) else { fatalError("Missing free text filter") }
```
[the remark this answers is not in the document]
shredlocker: Yea, I don't like that fatalError..

Sources/Charcoal/Filters/Root/RootFilterViewController.swift
shredlocker: I tried that first, but I think we need to scroll the inline filter even though it is not visible. It becomes visible after scroll the root view to top. Any ideas?

vadymmarkov: I think you can call `resetContentOffset` in `cellForRowAt` if `selectedItems` is empty. Or you can also try calling `resetContentOffset` in `prepareForReuse` in `InlineFilterCell`.

shredlocker: Should work fine now

shredlocker:
I'll make a new PR, looks like this one might break some stuff...

PR #208, opened 2019-03-27
Formatter numbers with thousand separator 

vadymmarkov:
# Why?

Because big numbers look better with separator.

# What?

- Add static number formatter to use across the app
- Format numbers with thousand separator everywhere except for years

# Show me

### Before

before

### After

after

### DIFF
--- a/Sources/Charcoal/Extensions/NumberFormatterExtensions.swift
+++ b/Sources/Charcoal/Extensions/NumberFormatterExtensions.swift
@@ -1,0 +1,21 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+extension NumberFormatter {
+    static let formatterWithSeparator: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .currency
+        formatter.currencySymbol = ""
+        formatter.locale = Locale(identifier: "nb_NO")
+        formatter.maximumFractionDigits = 0
+        return formatter
+    }()
+
+    func string(from value: Int) -> String? {
+        let number = NSNumber(value: value)
+        return string(from: number)
+    }
+}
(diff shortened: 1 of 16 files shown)


Sources/Charcoal/Extensions/NumberFormatterExtensions.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+extension NumberFormatter {
+    static let formatterWithSeparator: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .currency
```
shredlocker: I think we should use `.decimal`. I think the only difference is `.currency` has two decimals.

Sources/Charcoal/Filters/Root/RootFilterViewController.swift
```diff
@@ -111,6 +111,12 @@ final class RootFilterViewController: FilterViewController {
         ])
     }
 
+    private func updateBottomButtonTitle() {
+        let localizedString = String(format: "showResultsButton".localized(), filter.numberOfResults)
+        let title = localizedString.replacingOccurrences(of: "\(filter.numberOfResults)", with: filter.formattedNumberOfResults)
```
shredlocker: Nice 💯

Sources/Charcoal/Extensions/NumberFormatterExtensions.swift
vadymmarkov: Good point! Fixed.

Sources/Charcoal/Filters/Root/RootFilterViewController.swift
vadymmarkov: That's because `String(format:` removes whitespaces 🙈

PR #210, opened 2019-03-27
Scroll to top on reset button pressed

shredlocker:
# Why?

Should scroll to top on reset button pressed.

# What?

- Adds reset functionality to root and inline filters

# Show me

No UI

### DIFF
--- a/Sources/Charcoal/Filters/Root/RootFilterViewController.swift
+++ b/Sources/Charcoal/Filters/Root/RootFilterViewController.swift
@@ -45,6 +45,7 @@ final class RootFilterViewController: FilterViewController {
     }()
 
     private var freeTextFilterViewController: FreeTextFilterViewController?
+    private var indexPathsToReset: [IndexPath: Bool] = [:]
 
     // MARK: - Filter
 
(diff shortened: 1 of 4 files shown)


Sources/Charcoal/Filters/Root/RootFilterViewController.swift
```diff
@@ -123,8 +119,15 @@ final class RootFilterViewController: FilterViewController {
         selectionStore.removeValues(for: filter)
         rootDelegate?.rootFilterViewControllerDidResetAllFilters(self)
         freeTextFilterViewController?.searchBar.text = nil
+
+        for (index, subfilter) in filter.subfilters.enumerated() where subfilter.kind == .inline {
```
vadymmarkov: Do you get correct index from original `filter.subfilters` array here or an index from subarray `filter.subfilters.enumerated() where subfilter.kind == .inline`? 🤔

shredlocker: Good question! I don't know is the answer to that

bstien: You get the original index, seems like it checks each element before it provides it to the closure. Tried this one in a playground:

```swift
let a  = [1, 2, 3, 4, 5, 6]
for (index, elem) in a.enumerated() where elem % 2 == 0 {
    print(index) // 1, 3, 5
}
```

vadymmarkov: Cool 👍

PR #214, opened 2019-04-01
Accessibility: make Charcoal VoiceOver friendly

vadymmarkov:
# Why?

Because it should be possible to use `Charcoal` with VoiceOver.

# What?

All the following fixes were made for better VoiceOver experience:
- Step slider: fix increment and decrement
- Add accessibility values for units in range sliders and selection tags
- Hide bottom button when it's not visible on screen
- Set proper accessibility labels for list cells and verticals, make it clear which filters are selected
- Use better accessibility labels for selection tags
- Make it possible to scroll using VoiceOver when navigating through inline filters

# Show me

No UI changes.


Sources/Charcoal/Resources/en.lproj/Localizable.strings
```diff
@@ -19,22 +19,38 @@
...
+"unit.items.value" = "stk.";
+"unit.kilograms.value" = "kg";
+"unit.kilometers.value" = "km";
+"unit.seats.value" = "seter";
+"unit.squareMeters.value" = "m²";
+
+"unit.centimeters.accessibilityValue" = "centimeter";
+"unit.cubicCentimeters.accessibilityValue" = "kubikkcentimeter";
+"unit.currency.accessibilityValue" = "kroner";
+"unit.feet.value.accessibilityValue" = "fot";
```
shredlocker: Got an extra `value` in this one.

Sources/Charcoal/Resources/en.lproj/Localizable.strings
```diff
@@ -19,22 +19,38 @@
...
+"unit.horsePower.accessibilityValue" = "hestekrefter";
+"unit.items.accessibilityValue" = "stykker";
```
shredlocker: You might want to write `stykk` here.

vadymmarkov: Good that you checked Norwegian 😉

PR #215, opened 2019-04-02
Don't show context filter if it has only one subfilter

vadymmarkov:
# Why?

Because it doesn't affect the search results in any way if you select the only subfilter of a context filter.

# What?

Don't show context filter if it has only one subfilter.

# Show me

No UI changes.

### DIFF
--- a/Sources/FINNSetup/BackendModel/FilterSetup.swift
+++ b/Sources/FINNSetup/BackendModel/FilterSetup.swift
@@ -81,7 +81,7 @@ public struct FilterSetup: Decodable {
             return Filter.search(key: key)
         case FilterKey.preferences.rawValue:
             let subfilters = config.preferenceFilters.compactMap {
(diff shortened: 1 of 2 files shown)


Sources/FINNSetup/BackendModel/FilterSetup.swift
```diff
@@ -81,7 +81,7 @@ public struct FilterSetup: Decodable {
             return Filter.search(key: key)
         case FilterKey.preferences.rawValue:
             let subfilters = config.preferenceFilters.compactMap {
-                filterData(forKey: $0).map({ makeListFilter(from: $0, withStyle: .normal) })
+                filterData(forKey: $0).flatMap({ makeListFilter(from: $0, withStyle: .normal) })
```
shredlocker: Is this suppose to be here?

vadymmarkov: Yes, because `makeListFilter` returns optional now.

PR #225, opened 2019-04-04
Replace bottom UIButton with FinniversKit Button

shredlocker:
# Why?

Bottom button was too big. The way we had set up constraints could potentially be wrong with big fonts.

# What?

- Changed button from `UIButton` to `Button`
- Replaced constant button height with intrinsic content size
- Update constraints

# Show me

bottom-button


vadymmarkov:
Could you show how it looks and animates now?

PR #234, opened 2019-04-10
Bump to iOS 10

ag12:
# Why?
The main app uses iOS 10 and above.

# What?
Removed iOS 9 related code, and bumped target to 10.


Demo/MapFilter/DemoSearchLocationDataSource.swift
```diff
@@ -85,47 +62,3 @@ class DemoSearchLocationDataSource: NSObject, SearchLocationDataSource {
         return true
     }
 }
-
-@available(iOS 9.3, *)
-extension DemoSearchLocationDataSource: MKLocalSearchCompleterDelegate {
```
pimms: I'm struggling to see how this was being used...

I also don't understand exactly what `@available(ios 9.3, *)` means – how is it different from `@introduced`?

pimms: Oh, it was used here, duh:

```
            let localSearchCompleter = MKLocalSearchCompleter()
            localSearchCompleter.delegate = self
```

ag12: yes :)

Demo/MapFilter/DemoSearchLocationDataSource.swift
```diff
@@ -47,28 +35,17 @@ class DemoSearchLocationDataSource: NSObject, SearchLocationDataSource {
     ]
 
     func searchLocationViewController(_ searchLocationViewController: SearchLocationViewController, didRequestLocationsFor searchQuery: String, completion: @escaping ((SearchLocationDataSourceResult) -> Void)) {
-        if #available(iOS 9.3, *) {
```
vadymmarkov: I think it should be the other way around. We should keep this if statement because `#available(iOS 9.3, *)` means it is available on iOS10.

ag12: of course, you are right. :)

PR #237, opened 2019-04-11
Refactor demo project

shredlocker:
# Why?

Demo project wasn't really working after a lot of refactoring. This PR will clean up the project and make a demo for all components.

# What?

I'm not a big fan of using `enum` to represent static data like we did in the previous demo, so I experimented a bit with `UIViewController.Type` and using that to create view controllers lazily.

- Replaced demo view with view controllers.
- Updated FinniversKit

# Show me

No UI


Demo/ViewControllers/DemoTableViewCell.swift
```diff
@@ -0,0 +1,48 @@
...
+        selectedBackgroundView = view
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Public methods
+
+    func configure(with model: Row) {
```
vadymmarkov: I would just configure with title if it's the only thing needed from `Row` model.

Sources/Charcoal/CharcoalViewController.swift
```diff
@@ -57,6 +57,11 @@ public class CharcoalViewController: UINavigationController {
 
     // MARK: - Init
 
+    public override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
+        super.init(nibName: nibNameOrNil, bundle: nibBundleOrNil)
+        selectionStore.delegate = self
```
vadymmarkov: I think we can make `selectionStore` into lazy var and set delegate in closure. Then we don't have to override this init and will not forget to set delegate.

Demo/Config/Section.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import UIKit
+
+class Section: NSObject, UITableViewDataSource {
```
vadymmarkov: How about making `Section` into a simple struct with title and data only. Because `DataSource` is already `UITableViewDataSource` and I think it makes more sense to do table view setup there. What do you think?

Demo/Config/DataSource.swift
```diff
@@ -0,0 +1,66 @@
...
+    private lazy var sections = [
+        Section(title: "Components", data: components),
+        Section(title: "Markets", data: markets),
+    ]
+
+    // MARK: - Methods
+
+    func viewController(for indexPath: IndexPath) -> UIViewController {
+        let row = sections[indexPath.section].data[indexPath.row]
+        let viewController = row.type.init(nibName: nil, bundle: nil)
```
vadymmarkov: Can't we use `row.type.init()` here?

Demo/ViewControllers/DemoTableViewCell.swift
shredlocker: Yep

Demo/Config/Section.swift
shredlocker: I agree

Demo/Config/DataSource.swift
shredlocker: I think that should work

shredlocker:
I think we can merge this and then keep working on it as we go. Missing list and grid filter demo, and vertical selection is not working, but I want to work on that in a separate branch.

PR #238, opened 2019-04-12
Add a script to automate creation of tags

vadymmarkov:
# Why?

To make it easier to create tags according to semantic versioning from the command line.

# What?

Add a Swift script and compiled binary to automate creation of tags.

# Usage

- Show current version: `./version current`
- Create patch version: `./version patch -m 'Backward compatible bug fixes'`
- Create minor version: `./version minor -m 'Backward compatible new feature'`
- Create major version: `./version major -m 'Changes that break backward compatibility'`
- Delete last version: `./version delete`

You can also use `--force` is you want to skip confirmation.

**If we want to use this script it needs to be distributed/copied over to other repos we have. Also, not sure whether it's better/easier in comparison to doing it on GitHub, where it's more convenient to add description in markdown. Need you feedback on this.**


version.swift
```diff
@@ -0,0 +1,158 @@
...
+
+    func nextMajor() -> Tag {
+        return Tag(major: major + 1, minor: 0, patch: 0)
+    }
+}
+
+final class TagManager {
+    func loadCurrentTag() -> Tag? {
+        let regex = "^[0-9]*\\.[0-9]*\\.[0-9]"
+        let gitTag = execute("git tag | sort -r | grep '\(regex)' | head -1 2> /dev/null") ?? ""
```
bstien: I don't think `sort -r` does numeric sorting. I.e. given this list of versions: `4.1 4.2 … 4.10 4.10.1` it would be sorted as:

```
4.2
4.10.1
4.10
4.1
```

vadymmarkov: I think you're right

vadymmarkov: Should be fixed now.

version.swift
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    func nextMajor() -> Tag {
+        return Tag(major: major + 1, minor: 0, patch: 0)
+    }
+}
+
+final class TagManager {
+    func loadCurrentTag() -> Tag? {
+        let regex = "^[0-9]*\\.[0-9]*\\.[0-9]"
```
bstien: I think you would also need a complete stop, aka. `$`, at the end of this regex. Just in case someone names their tag `4.49.1-rc1`.

bstien: Could potentially also specify that you need at least 1 digit in each segment.

I.e.
```swift
let regex = "^\\d+\\.\\d+\\.\\d+$"
```

vadymmarkov: Grep's syntax doesn't support notations like \d. But I changed it to
```swift
let regex = "^[0-9]\\+\\.[0-9]\\+\\.[0-9]\\+$"
```

bstien: Ah, right. For that to be allowed you need to add the `-E` argument, aka. extended regex.

But the one you changed to is more than sufficient 👌

PR #241, opened 2019-04-12
Remove custom fonts in favor of fonts from FinniversKit

vadymmarkov:
# Why?

Because we should use fonts from `FinniversKit` instead of creating custom fonts from the same resource files.

# What?

- Remove font resources and unused extensions
- Adapt fonts from `FinniversKit` to custom font sizes we use in some places

# Show me

That's the bigger UI change, before and after:

Screenshot 2019-04-12 at 15 02 22


Sources/Charcoal/Filters/Range/Input/NumberInputView.swift
```diff
@@ -257,20 +257,20 @@ extension NumberInputView: UITextFieldDelegate {
 private struct Style {
     static let textColor: UIColor = .licorice
     static let errorTextColor: UIColor = .cherry
-    static let hintNormalFont: UIFont? = UIFont(name: FontType.light.rawValue, size: 16)
-    static let hintActiveFont: UIFont? = UIFont(name: FontType.medium.rawValue, size: 16)
+    static let hintNormalFont: UIFont = .body
+    static let hintActiveFont: UIFont = .bodyStrong
```
bstien: Just a hint, the names _might_ change yet again😞 Got some super secret inside info from @osanoj that the designers are discussing the namings.

PR #242, opened 2019-04-15
Removed some iOS 9 leftovers

ag12:
# Why?
The main app has bumped to iOS 10. 

# What?
Removed iOS 9 leftover code.


Demo/MapFilter/DemoSearchLocationDataSource.swift
```diff
@@ -17,15 +17,11 @@ class DemoSearchLocationDataSource: NSObject, SearchLocationDataSource {
     private var localSearch: MKLocalSearch?
 
     private lazy var localSearchCompleter: NSObject? = {
```
shredlocker: You can replace `NSObject?` with `MKLocalSearchCompleter` I guess ;)

ag12: Yes we can 😄

PR #262, opened 2019-05-02
Removing lease related filters from car abroad

osanoj:
# Why?
Some filters that should not be there.

# What?
Removed them from car-abroad config.

# Show me

### Before
Simulator Screen Shot - iPhone 8 - 2019-05-02 at 10 15 26

### After
Simulator Screen Shot - iPhone 8 - 2019-05-02 at 10 16 59


vadymmarkov:
Shouldn't these appear as context filters?

osanoj:
Not on cars abroad apparently

PR #269, opened 2019-05-06
Enable swipe back navigation in range and map view controllers

vadymmarkov:
# Why?

Because it's natural to have swipe back navigation for every view controller pushed into navigation stack.

# What?

- Detect left edge swipes and disable `interactivePopGestureRecognizer` if touch received from an instance of `UISlider`
- Enable `interactivePopGestureRecognizer` on `viewWillAppear` and `viewWillDisappear` in case it's disabled after using the slider

# Show me

swipe_back

### DIFF
--- a/Sources/Charcoal/Filters/FilterViewController.swift
+++ b/Sources/Charcoal/Filters/FilterViewController.swift
@@ -78,23 +78,7 @@ class FilterViewController: UIViewController, FilterBottomButtonViewDelegate {
         enableSwipeBack(true)
     }
 
-    // MARK: - Setup
-
-    private func setup() {
-        view.addSubview(topSeparatorView)
-        view.addSubview(bottomButton)
-
-        NSLayoutConstraint.activate([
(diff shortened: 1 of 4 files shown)


Sources/Charcoal/Filters/FilterViewController.swift
```diff
@@ -60,31 +60,26 @@ class FilterViewController: UIViewController, FilterBottomButtonViewDelegate {
         view.backgroundColor = .milk
         setup()
         hideTopSeparator()
+
+        let gestureRecornizer = UIScreenEdgePanGestureRecognizer()
```
shredlocker: Typo here

vadymmarkov:
As a side note, setting `interactivePopGestureRecognizer?.delegate` didn't work at all. Instead of sending events to its delegate it just disabled swipe back gesture entirely.

shredlocker:
Good job Apple! 😆

PR #273, opened 2019-05-07
Free text filter: fix keyboard hiding suggestion rows

vadymmarkov:
# Why?

Because it wasn't possible to see all table view cells due to keyboard being presented.

# What?

Set content inset when keyboard is being presented.

# Show me

### Before

before

### After

after

### DIFF
--- a/Sources/Charcoal/Filters/FreeText/FreeTextFilterViewController.swift
+++ b/Sources/Charcoal/Filters/FreeText/FreeTextFilterViewController.swift
@@ -54,11 +54,11 @@ public class FreeTextFilterViewController: UIViewController {
 
     private lazy var tableView: UITableView = {
         let tableView = UITableView(frame: .zero, style: .plain)
+        tableView.translatesAutoresizingMaskIntoConstraints = false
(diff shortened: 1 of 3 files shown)


Sources/Charcoal/Filters/FreeText/FreeTextFilterViewController.swift
```diff
@@ -53,34 +54,71 @@ public class FreeTextFilterViewController: UIViewController {
...
 
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
 
+    // MARK: - Lifecycle
+
+    public override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        let notificationCenter = NotificationCenter.default
```
bstien: I guess you could remove this line, since you get a `NotificationCenter` through the constructor.

vadymmarkov: Nice catch!

PR #277, opened 2019-05-09
Grid filter: fix selection

vadymmarkov:
# Why?

Because filters won't toggle if you re-enter the grid filter view.

# What?

Select filter using `selectItem(at:` method from `UICollectionView`.

# Show me

### Before

before

### After

after


vadymmarkov:
Strange indeed.

PR #279, opened 2019-05-09
Invert map radius overlay colors

vadymmarkov:
# Why?

Because it makes it easier to clearly see the selected area.

# What?

Invert map radius overlay colors.

# Show me

map


vadymmarkov:
Now it looks even better with corner radius:

corners

PR #282, opened 2019-05-09
Add shadow to bottom button

shredlocker:
# Why?

We want shadows at the bottom as well as the top.

# What?

- Added shadow to `FilterBottomButtonView`
- Added `scrollView` variable to `ScrollViewController` class in order to set correct shadow in `viewWillAppear`
- All shadows follows scrolling to make the transition smoother.
- `SearchLocationViewController` now extends `ScrollViewController` and will also have shadow beneath the search bar
- Made `topSeperatorView` as tall as the navigation bar by default with the possibility to increase the height when used with a search bar.
- Gave `InlineFilter` a bit more padding above and below.

# Show me

shadows


Sources/Charcoal/Filters/FilterViewController.swift
```diff
@@ -59,19 +63,26 @@ class FilterViewController: ScrollViewController, FilterBottomButtonViewDelegate
         super.viewWillAppear(animated)
         view.bringSubviewToFront(bottomButton)
         enableSwipeBack(true)
+        updateBottomShadow()
```
vadymmarkov: Why do we need to do this on `viewWillAppear`?

Sources/Charcoal/Filters/List/ListFilterViewController.swift
```diff
@@ -33,6 +33,7 @@ final class ListFilterViewController: FilterViewController {
     init(filter: Filter, selectionStore: FilterSelectionStore) {
         self.filter = filter
         super.init(title: filter.title, selectionStore: selectionStore)
+        scrollView = tableView
```
vadymmarkov: Is it possible to avoid setting this in `init` because it breaks lazy initialization of `tableView`?

Sources/Charcoal/Filters/FilterViewController.swift
shredlocker: We have to check the content size and see if we need a shadow before scrolling.

Sources/Charcoal/Filters/List/ListFilterViewController.swift
shredlocker: I'll see if I can move it somewhere else

shredlocker:
What to you think about this @vadymmarkov ?

vadymmarkov:
Looks good! But can't we still have `ScrollViewController` where `shadowView` is created and updated on `scrollViewDidScroll`?

shredlocker:
Yep, I'll change that ;)

PR #286, opened 2019-05-19
Range number input: fix cursor position and text replacement 

vadymmarkov:
# Why?

Because it wasn't possible to remove whitespaces and the cursor position was constantly moved to the end while editing text.

# What?

Take whitespaces into account and keep cursor position while editing text.

# Show me

### Before

before

### After

after


Sources/Charcoal/Extensions/StringExtensions.swift
```diff
@@ -6,4 +6,29 @@ extension String {
...
+    func removingWhitespaces() -> String {
+        let components = self.components(separatedBy: .whitespaces)
+        return components.joined(separator: "")
+    }
+
+    mutating func removeWhitespaces() {
+        self = removingWhitespaces()
+    }
+
+    func range(from range: NSRange, replacementString: String) -> Range<String.Index>? {
```
osanoj: Can this be named to someway describe more what it does? Since the whole removing whitespace isn't exactly standard.

UnitTests/Charcoal/Extensions/StringExtensionsTests.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+@testable import Charcoal
+import XCTest
+
+final class StringExtensionsTests: XCTestCase {
```
osanoj: 👏

Sources/Charcoal/Extensions/StringExtensions.swift
vadymmarkov: Do you have any good names in mind? Perhaps `rangeConsideringWhitespaces` or something like that?

osanoj: Yes, maybe something like that, then it is obvious you use it for "special" things (avoid using this by mistake somewhere else). Or is it `replacementRangeNotConsideringWhitespaces` 🤔 are we considering the whitespaces or skipping them? 😂 

`rangeWithSpecialWhitespaceHandling`? 😛

vadymmarkov: This func makes a range bigger so both the whitespace and the number before it will be removed. But I think `replacementRangeNotConsideringWhitespaces` should work.

vadymmarkov: Renamed 😉

PR #293, opened 2019-06-26
Updated FinniversKit

pimms:
I added some delegate methods in the latest PR to FinniversKit. The protocol was not implemented in Charcoal, but we should probably still stay up to date to avoid conflicts when integrating all of this in the FINN app.


pimms:
Disregard this for a little bit, I forgot to update the version in FK.

PR #295, opened 2019-08-14
Add anchor for popoverPresentationController on iPad

hakloev:
# Why?

Need to be able to present the `UIAlertController` for clearing the filter on iPad.

# What?

Anchored the `popoverPresentationController` to the `UIBarButtonItem` `resetButton` if it's present.

# Show me

### Before

Was not available on the iPad

### After

Simulator Screen Shot - iPad Air - 2019-08-14 at 10 05 53


3lvis:
📱

PR #297, opened 2019-08-22
Feature: CalloutView for bottomButton

hakloev:
# Why?

On iPad the filter will be presented in a drawer, and thus we need the `CalloutView` to ensure that the users sees the "Bruk"-button to activate their selected filters.

# What?

* Updated `FinniversKit` to version 3.2.0
* Added a `BottomButtonCalloutOverlay` similar to `VerticalCalloutOverlay`
* Added `filterViewcontrollerWillPresentBottomButton(viewController:)` to `FilterViewControllerDelegate` in order to notify `CharcoalViewController` that the bottom button is becoming visible, and that the callout maybe should become visible.
* Added a view controller for demoing how it looks on iPad

# Show me

### Before

Wasn't available

### After

Simulator Screen Shot - iPad Air - 2019-08-22 at 08 36 39

Simulator Screen Shot - iPad Air - 2019-08-22 at 08 46 48


Sources/Charcoal/Filters/Root/Views/BottomButtonCalloutOverlay.swift
```diff
@@ -0,0 +1,72 @@
...
+        backgroundColor = .clear
+
+        addSubview(bodyView)
+        addSubview(calloutView)
+
+        let gestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(handleTap))
+        addGestureRecognizer(gestureRecognizer)
+
+        NSLayoutConstraint.activate([
+            calloutView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -76),
```
hakloev: I'm not happy with this magic constant that is equal to the height of the `FilterBottomButtonView` height. If there are any tips on how to get the height I'll gladly take them 🙃 One solution I tried out were to get the height from the `intrinsicContentSize` on `FilterBottomButtonView` when making the bottom button visible and pass it through the `configure(withText:)` method, but it felt dirty 🤔

shredlocker: I guess you can constrain it to the `topAnchor` of the bottomButton. You can pass an anchor to the configure method f.eks.
`configure(withText text: String, anchor: NSLayoutYAxisAnchor)`

shredlocker:
Maybe you can replace both of these callouts with a single callout view? Create a model with text and direction, and then use the configure method to anchor it to the correct view.

hakloev:
I've made an attempt on what you commented on now, but with a slightly different approach, since I needed to set the direction on the CalloutView during initialisation.

PR #298, opened 2019-08-23
Drop iOS 10

fespinoza:
# Why?

As we are updating all our libraries to iOS 11.2+

# What?

- update the iOS deployment version for all targets

# Show me

No UI changes


fespinoza:
One thing i am not entirely sure here is about the FinniversKit version Charcoal depends on as that version is not specified in the podspec

pimms:
I would think that Charcoal doesn't specify a FinniversKit version to minimize potential conflicts when integrating in `ios-app` — I believe we're doing something similar in the podspec of the Messaging SDK.

fespinoza:
That's true, in this case Charcoal also needed a fix for a breaking change in `FinniversKit`

PR #301, opened 2019-09-17
Increase default maxPrice from 10 mill to 15 mill

bstien:
# Why?
We want to increase the max price for realestate (BTS) to 15 millions, instead of 10. Ref. a task on our Trello board.

# What?
- Increased `maximumValue`.


osanoj:
We couldn't also add variable increments too? Would be very nice 😄

bstien:
Good idea, I'll talk to Solfrid💯

bstien:
Discussed with Thuy about the increment. Keeping it at 100k for now, instead of variable increments.

Ready for review again :)

PR #303, opened 2019-10-04
Charcoal demo 5.1.1

osanoj:
# Why?
We need to move from distributing the Demo with HockeyApp to App Center. This is first step towards that.

This is now still using the HockeySDK but the app has been migrated to App Center. Next step is to switch SDK and see what happens 😄 

# What?
The main thing I had to do in order to build the beta with fastlane was in commit https://github.com/finn-no/charcoal-ios/commit/41c38c83dda226eeaf64a415080a468de8d824c2 (_Made the Demo target build as 'Debug' even when archived because otherwise a build to distribute could not handle all the '@testable import Charcoal'_). I don't think we used '@testable import Charcoal' when the last beta was made. 

# Show me
Screen Shot 2019-10-04 at 14 28 05


3lvis:
😻 😻 😻 😻

PR #304, opened 2019-10-22
Feature: 🎅Xmas filters🎁

bstien:
# Why?
We need to add support for the new xmas ad types on Torget.

I encountered a couple of issues when making this PR. The actual changes needed to support xmas filters are 2 lines, but since our platform provider has made changes to spacings and `UITableView` behavior in iOS 13 this took some more work.

## UITableView
Initially, this built fine on iOS 11.4, but I got a blank tableView when I entered a `ListFilterViewController`.

I also got this warning in the console, saying that we tried to layout the tableView without having it in the view hierarchy.

> Warning once only: UITableView was told to layout its visible cells and other contents without being in the view hierarchy (the table view or one of its superviews has not been added to a window). 
> This may cause bugs by forcing views inside the table view to load and perform layout without accurate information (e.g. table view bounds, trait collection, layout margins, safe area insets, etc), and will also cause unnecessary performance overhead due to extra layout passes.
> 
> Make a symbolic breakpoint at **UITableViewAlertForLayoutOutsideViewHierarchy** to catch this in the debugger and see what caused this to occur, so you can avoid this action altogether if possible, or defer it until the table view has been added to a window. 

This happens when we try to present or remove the `bottomButton` before we push the new viewController into the navigationStack. The culprit is the call to `view.layoutIfNeeded()`.

To prevent this I've added a check to make sure to only call this method if the view is on screen.

## Spacings
iOS 13 has changes to spacings, which I've tried to fix/explain in this PR: https://github.com/finn-no/FinniversKit/pull/667.

# What?
- Present `christmas` filter for BAP.
- Only call `view.layoutIfNeeded()` if view is on screen.
- Change constraint constant for cells that has a chevron.
- Update FinniversKit.
- Update CircleCI config.

# Show me

## Christmas filter

| List | Selection |
| --- | --- |
| Screenshot 2019-10-22 at 14 16 31 | Screenshot 2019-10-22 at 14 16 33 |

## Spacing changes

| iOS version| Before | After |
| --- | --- | --- |
| iOS 13 | Screenshot 2019-10-22 at 12 48 05 | Screenshot 2019-10-22 at 13 19 23 |
| iOS 11.4 | Screenshot 2019-10-22 at 13 07 10 | Screenshot 2019-10-22 at 13 22 42 |


Sources/Charcoal/Filters/FilterViewController.swift
```diff
@@ -72,7 +72,10 @@ class FilterViewController: ScrollViewController, FilterBottomButtonViewDelegate
     // MARK: - Internal functions
 
     func showBottomButton(_ show: Bool, animated: Bool) {
-        view.layoutIfNeeded()
+        if viewIfLoaded?.window != nil {
```
vadymmarkov: It's weird we have to do this 😅

Sources/Charcoal/Filters/Root/Views/RootFilterCell.swift
```diff
@@ -12,7 +12,7 @@ protocol RootFilterCellDelegate: AnyObject {
 final class RootFilterCell: BasicTableViewCell {
     weak var delegate: RootFilterCellDelegate?
 
-    var isEnabled = true {
+    override var isEnabled: Bool {
```
vadymmarkov: We need this change in order to update `FinniversKit` in the iOS app 👍

Sources/Charcoal/Filters/FilterViewController.swift
bstien: Agree..!😅😅😅

fespinoza:
> Awesome 👍 We should probably use SPM instead of Carthage like we did in FinniversKit.

I imagine we cannot use SPM because finniversKit has assets, which are not supported by SPM

vadymmarkov:
This is true, and I think we also use some FinniversKit assets in Charcoal.

PR #308, opened 2019-10-31
Dark mode support

fespinoza:
# Why?

_hello darkness my old friend_...

# What?

- Use semantic colors instead of concrete ones
- Add `Theme.mainBackground` as the background color for the charcoal components is `.darkIce` in dark mode and `.milk` in light mode
- Update FinniversKit thought carthage

# Show me

| Light | Dark |
| --- | --- |
| inline-light | inline-dark |
| main-light | main-dark |
| map-light | map-dark |
| onboarding-light | onboarding-dark |
| range-light | range-dark |

| Before | After |
| --- | --- |
| torget-before |  torget-after |


Sources/Charcoal/Theme.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+
+import FinniversKit
+import UIKit
+
+public class Theme {
+    /// The background used for charcoal will be different in dark mode than the standard FinniversKit as charcoal
+    /// components are presented on top of other elements, then in order to achieve more contrast and a sense of levels
+    /// we need to override the value for the main background color for dark mode.
+    public class var mainBackground: UIColor {
```
shredlocker: Do we not have a bottom sheet color in FinniversKit? I think we want to use this color for all bottom sheets, or am I wrong?

fespinoza: indeed, the idea is to use this color for bottom sheets so it can be added in FinniversKit instead. 

Although I would probably keep the mainBackground property as i feel it fits better for Charcoal than set background to `button = bottomSheetBackground` for example. Then this color would reference the bottom sheet one

PR #312, opened 2019-11-27
Fix: avoid automatic re-centering on the filter map after scrolling

fespinoza:
# Why?

@hakloev reported a bug about this https://sch-chat.slack.com/archives/G0903G630/p1574771877180300

The functionality was removed in #310 and caused that the map would re-center on the user
location after scrolling away. Curiously enough, this only seem to happen on device, not on the simulator

# What?

- Restore `hasRequestedLocationAuthorization`


osanoj:
I guess this also reintroduces the issue that #310 fixed?

fespinoza:
> I guess this also reintroduces the issue that #310 fixed?

PR #316, opened 2020-01-10
Don't show root filter with no selectable values

osanoj:
# Why?
A root filter is not useful if it has no subfilters.
 
# What?
Removed root filters without queries (that are not `map`, `shoeSize` or a range).

# Show me
This is how it effects the main app, look at "Juleglede"

### Before
Simulator Screen Shot - iPhone 11 Pro - 2020-01-10 at 11 16 23

### After
Simulator Screen Shot - iPhone 11 Pro - 2020-01-10 at 11 11 47

### DIFF
--- a/Sources/FINNSetup/BackendModel/FilterSetup.swift
+++ b/Sources/FINNSetup/BackendModel/FilterSetup.swift
@@ -97,6 +97,10 @@ public struct FilterSetup: Decodable {
                     return nil
                 }
             } else {
+                guard data.queries.count > 0 else {
+                    // No need to show a root filter that has no selectable values
+                    return nil
+                }
                 return makeFilter(from: data, withKind: .standard, style: style)
             }
         }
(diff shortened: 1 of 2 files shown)


Sources/FINNSetup/BackendModel/FilterSetup.swift
```diff
@@ -97,6 +97,10 @@ public struct FilterSetup: Decodable {
                     return nil
                 }
             } else {
+                guard data.queries.count > 0 else {
```
shredlocker: I would probably use `isEmpty` here instead of `count`.
I'm not sure if the comment below is necessary in this situation, but it doesn't hurt to leave it 😄

osanoj: I agree 👍  

Removed the comment too, it should be understandable from just the code

PR #317, opened 2020-01-10
WIP: Feature toggle support suggestion/discussion

osanoj:
# Why?
We have had a couple of situations now where we want to toggle some feature based on unleash, but since we do not want to access unleash from within Charcoal it has been a bit complicated.

# What?
This is a suggestion for how we could have a feature toggle solution for Charcoal. It is a bit extra complicated since we have uses for it both in Charcoal core and our FINN setup part. I did not find a natural place to "hand it over" to Charcoal either but put it as a property in `FilterSetup` for now.

Free to come with better (or just different) suggestions for how to solve this. What I implemented probably needs at least some refactoring (and renaming) before it is ready to use (for instance `bottomButtonCallout` "feature" is not implemented yet).

If we where to do it this way the main app needs to have implementations for `FeatureConfig` and `FeatureInfo` that connects to unleash toggles (and supplies text for region reform callout). It would also need to add writing to `UserDefaults` for those things that are to be shown only once.

### DIFF
--- a/Sources/Charcoal/CharcoalViewController.swift
+++ b/Sources/Charcoal/CharcoalViewController.swift
@@ -69,11 +69,11 @@ public final class CharcoalViewController: UINavigationController {
     public override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
 
-        let userDefaults = UserDefaults.standard
-
-        if let text = filterContainer?.regionReformCalloutText, !userDefaults.regionReformCalloutShown {
+        if let regionReformFeature = filterContainer?.featureConfig?.featureConfig(.regionReformCallout),
+            regionReformFeature.isEnabled,
+            let text = regionReformFeature.text {
             showCalloutOverlay(withText: text, andDirection: .down, constrainedToTopAnchor: navigationBar.bottomAnchor)
-            userDefaults.regionReformCalloutShown = true
+            regionReformFeature.didShow()
         }
     }
 
@@ -333,11 +333,6 @@ extension CharcoalViewController: CalloutOverlayDelegate {
 // MARK: - UserDefaults
 
 private extension UserDefaults {
-    var regionReformCalloutShown: Bool {
-        get { return bool(forKey: "Charcoal." + #function) }
-        set { set(newValue, forKey: "Charcoal." + #function) }
-    }
-
     var bottomButtomCalloutShown: Bool {
         get { return bool(forKey: #function) }
         set { set(newValue, forKey: #function) }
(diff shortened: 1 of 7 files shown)


vadymmarkov:
In general, it looks good to me, but I think we don't have any callout text to show for Finn features, so this could be split into two protocols: 

```swift
public protocol CharcoalFeatureConfig {
   func isEnabled(_ feature: CharcoalFeature)
   func calloutText(for feature: Feature)
   func didShow(_ feature: Feature)
}

public protocol FINNFeatureConfig {
  func isEnabled(_ feature: FINNFeature) 
}
```

As for Charcoal features, do we want to control them with Unleash as well?

Sources/Charcoal/CharcoalViewController.swift
```diff
@@ -333,11 +333,6 @@ extension CharcoalViewController: CalloutOverlayDelegate {
 // MARK: - UserDefaults
 
 private extension UserDefaults {
-    var regionReformCalloutShown: Bool {
-        get { return bool(forKey: "Charcoal." + #function) }
```
vadymmarkov: We should probably be careful with this, because by changing the key in user defaults we'll show the callout for those who have already seen it.

osanoj: Yes, that definitely needs to be double checked for all things saved in user defaults before making a change like this

osanoj:
> In general, it looks good to me, but I think we don't have any callout text to show for Finn features, so this could be split into two protocols:
> 
> ```swift
> public protocol CharcoalFeatureConfig {
>    func isEnabled(_ feature: CharcoalFeature)
>    func calloutText(for feature: Feature)
>    func didShow(_ feature: Feature)
> }
> 
> public protocol FINNFeatureConfig {
>   func isEnabled(_ feature: FINNFeature) 
> }
> ```

Yes we could do that and only add it if we need it later. I was also thinking of having a "base" protocol and then extending it with "capabilities", i.e. base would only have `isEnabled`. Something like:

```
protocol FeatureConfig {
  func isEnabled(_ feature: CharcoalFeature)
}

protocol TextFeature {
  var text { get }
}

protocol ShownUpdatingFeature {
  func didShow(_ feature: Feature)
}
```

But not sure if it is worth it for the complexity it adds.


> 
> As for Charcoal features, do we want to control them with Unleash as well?

What I'm thinking now is that Charcoal does not need to know if the feature is controlled with unleash/date/beta-build or a combination. It just needs to know if something is currently enabled or not.

vadymmarkov:
I think if we're splitting "features" into two "CharcoalFeature" and "FINNFeature" enums, it's easier with two independent `CharcoalFeatureConfig` and `FINNFeatureConfig` protocols. Protocols with extended "capabilities" is also a good idea, but then we'll probably end up with multiple configuration properties in `FINNSetup` like `textFeatures`, `shownUpdatingFeatures` to avoid type casting.

osanoj:
> I think if we're splitting "features" into two "CharcoalFeature" and "FINNFeature" enums, it's easier with two independent `CharcoalFeatureConfig` and `FINNFeatureConfig` protocols. Protocols with extended "capabilities" is also a good idea, but then we'll probably end up with multiple configuration properties in `FINNSetup` like `textFeatures`, `shownUpdatingFeatures` to avoid type casting.

Yes, I'm a bit afraid we will end up with all possible combinations `textFeatures`, `shownUpdatingFeatures`, `textAndShownUpdatingFeatures` etc, therefore I put it all into the same `CharcoalFeatureConfig` despite them not being used for everything.

PR #329, opened 2020-03-17
Corona aid

osanoj:
# Why?
To make it possible to have corona aid as a separate filter on root level.

# What?
- Whitelisted "corona-aid".
- Added possibility to exclude some filters from root level when building a FilterContainer. So main app can use unleash to decide if this is to be at root level or not.

# Show me
Have a look at PR in main app.


osanoj:
Hmm, tests doesn't build, but see it is the same for latest commit to master too 🤔

osanoj:
Tests run fine on my machine so I'll push this through despite CircleCI, but someone™️ should have a look at that issue soon 😄

PR #332, opened 2020-04-23
Polygon search

ninarg:
# Why?

Today, users are restricted to search for items based on county, city or a radius distance. Some users are complaining about the new counties being too big. The polygon search gives the users more flexibility, as it allows you to customise your own area. 

# What?

Renamed `MapFilterViewController` to `MapRadiusFilterViewController`. Added new VC `MapPolygonFilterViewController`. `MapTabBarController` toggles them. 

Apple has had some MapKit bugs recently. On iOS 13.2 and 13.3 there was a bug that caused overlays to flicker every time you added/removed one, therefore I will disable polygon search for these versions. 

The bug was fortunately to some extent fixed in iOS 13.4. But it’s still not as good as it used to be. I have been testing with my iPhone 7 with iOS 11, and the animations are perfect there. In iOS 13.4, you are able to see some lag as the polygon is being rendered. I think this is out of our control, and did not find a fix for it, but I'm happy to hear if you have any suggestions. I still think it looks good enough to be released for iOS 13.4.

In the table below I added demos so you can see the rendering difference.

| iOS 11 | iOS 13.4 |
| --- | --- |
| polygon-iphone7 | polygon-ios13 |

# Show me

polydemo

| Feature | Description |
| --- | --- | 
| IMG_1692 | The user initially has to choose a place to attach the polygon, before the polygon annotations appear. There's an overlay over the map, where you can zoom/drag. | 
| IMG_1691 | When using location search, the user is asked if they want to keep the polygon where it was, or reset it. | 
| IMG_1693 | If user clicks reset button, they're asked if they really wanna delete the polygon. | 
| IMG_1694 | After user reached max number of points, an info view appears. | 
| IMG_1679 | When the info view above disappears, a callout explains polygon points can be deleted. |


Sources/Charcoal/Filters/Map/Polygon/PolygonEdge.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    func intersects(with edge: PolygonEdge) -> Bool {
+        return intersect(p1: p1, p2: p2, q1: edge.p1, q2: edge.p2)
+    }
+
+    private func intersect(p1: CGPoint, p2: CGPoint, q1: CGPoint, q2: CGPoint) -> Bool {
+        return ccw(p1, q1, q2) != ccw(p2, q1, q2) && ccw(p1, p2, q1) != ccw(p1, p2, q2)
+    }
+
+    private func ccw(_ p1: CGPoint, _ p2: CGPoint, _ p3: CGPoint) -> Bool {
```
fespinoza, line 24: can you add a more descriptive name here?

Sources/Charcoal/Filters/Map/Polygon/Polygon.swift
```diff
@@ -0,0 +1,48 @@
...
+import Foundation
+
+class Polygon {
+    private var edges: [PolygonEdge]
+
+    init(edges: [PolygonEdge]) {
+        self.edges = edges
+    }
+
+    func hasIntersectingEdges() -> Bool {
```
fespinoza, line 14: nitpick, this would be better as a computed variable

Sources/Charcoal/Filters/Map/Polygon/Views/InfoView.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    private lazy var boxView: UIView = {
+        let view = UIView(withAutoLayout: true)
+        view.backgroundColor = .bgPrimary
+
+        view.layer.shadowColor = UIColor.black.cgColor
+        view.layer.shadowOffset = CGSize(width: 0, height: 1)
+        view.layer.shadowRadius = 5
+        view.layer.shadowOpacity = 0.2
+        view.layer.cornerRadius = 16
```
fespinoza, line 19: `.spacingM`

Sources/Charcoal/Filters/Map/Polygon/Views/InfoView.swift
```diff
@@ -0,0 +1,71 @@
...
+import Foundation
+
+class InfoView: UIView {
+    // MARK: - Subviews
+
+    private lazy var boxView: UIView = {
+        let view = UIView(withAutoLayout: true)
+        view.backgroundColor = .bgPrimary
+
+        view.layer.shadowColor = UIColor.black.cgColor
```
fespinoza, line 15: there is a `dropShadow` helper in FinniversKit

Sources/Charcoal/CharcoalViewController.swift
```diff
@@ -250,6 +262,17 @@ extension CharcoalViewController: FilterViewControllerDelegate {
...
+
+    private func pushTabBarController(_ controller: UITabBarController) {
+        controller.hidesBottomBarWhenPushed = true
+        pushViewController(controller, animated: true)
+    }
+
+    private func showPolygonSearchCalloutIfNeeded() {
+        guard !UserDefaults.standard.polygonSearchCalloutShown else { return }
+        UserDefaults.standard.polygonSearchCalloutShown = true
+        showCalloutOverlay(withText: "map.polygonSearch.callout.title".localized(), andDirection: .up, andArrowAlignment: .right(15), constrainedToDirectionalAnchor: navigationBar.bottomAnchor)
```
fespinoza, line 274: in cases where methods get too long horizontally, it's always nice to split them in several lines, it's much more readable. This applies in several parts of this PR, which is not that critical but more a style preference (at least for me)

Sources/Charcoal/CharcoalViewController.swift
```diff
@@ -342,4 +373,9 @@ private extension UserDefaults {
         get { return bool(forKey: #function) }
         set { set(newValue, forKey: #function) }
     }
+
+    var polygonSearchCalloutShown: Bool {
+        get { return bool(forKey: #function) }
```
osanoj, line 378: Please prefix this with "Charcoal." like `regionReformCalloutShown`, all userdefaults usage should be prefixed when in a framework. Unfortunately we made a mistake with that for `bottomButtomCalloutShown `.

Sources/Charcoal/Filters/Map/MapTabBarController.swift
```diff
@@ -0,0 +1,121 @@
...
+//  Copyright © FINN.no AS, Inc. All rights reserved.
+//
+
+import Foundation
+
+protocol MapTabBarControllerDelegate: AnyObject {
+    func mapTabBarController(_ mapTabBarController: MapTabBarController, didSelect selection: CharcoalViewController.PolygonSelection)
+}
+
+class MapTabBarController: UITabBarController {
```
osanoj, line 11: I'm a bit unsure about using a `UITabBarController` here, usually you get issues when it is not in the root of the hierarchy, but maybe it works here because you are not actually using the tab view... Is there any reason for subclassing `UITabBarController` over using a normal `UIViewController` with `children`?

Sources/Charcoal/CharcoalViewController.swift
```diff
@@ -293,6 +316,14 @@ extension CharcoalViewController: UINavigationControllerDelegate {
     }
 }
 
+// MARK: - MapTabBarControllerDelegate
+
+extension CharcoalViewController: MapTabBarControllerDelegate {
+    func mapTabBarController(_ mapTabBarController: MapTabBarController, didSelect selection: CharcoalViewController.PolygonSelection) {
+        selectionDelegate?.charcoalViewController(self, didSelect: selection)
```
osanoj, line 323: I'm not quite sure why we need a specialized `didSelect` method for polygon? Because there is no similar for the radius filter being selected? Or maybe I just didn't see that :) 

Perhaps it will be clear once I see how it is integrated.

Sources/Charcoal/Filters/FilterViewController.swift
```diff
@@ -125,7 +126,20 @@ public class FilterViewController: ScrollViewController, FilterBottomButtonViewD
 
 extension FilterViewController: UIGestureRecognizerDelegate {
     public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldReceive touch: UITouch) -> Bool {
+        if let view = gestureRecognizer.view as? MKAnnotationView,
+            view.reuseIdentifier == MapPolygonFilterViewController.annotationId {
```
osanoj, line 130: Did you consider subclassing `MKAnnotationView` to a `PolygonPinAnnotationView` so you could check if it was an object of that type instead of looking at the `reuseIdentifier`?

Sources/Charcoal/CharcoalViewController.swift
ninarg: It's basically just for tracking. I added this enum, so that I don't have to add a new function for everything I want to track in the map search. Maybe we should just have a more generic `TrackingEvent` instead? Or call it something else.

osanoj: Aha, ok, then I understand. 

Probably fine to do this, we should maybe change to something generic if we need it for more things.

Sources/Charcoal/Filters/FilterViewController.swift
ninarg: There is no polygon view, I just use `MKAnnotationView`. But I will do this instead: `view.annotation is PolygonSearchAnnotation`, so that I don't have to use the `reuseIdentifier` 👍

Sources/Charcoal/Filters/Map/MapTabBarController.swift
ninarg: There wasn't really a benefit to use `UITabBarController` when I had another look. So now I changed it into a `UIViewController` called `MapFilterViewController`. I also changed so that `MapFilterViewController` subclasses `FilterViewController` (instead of radius and polygon VC doing that), so I handle the bottom button + toggling there, and keep Radius+Polygon as child VC in a `mapContainerView`. You can have another look!

Sources/Charcoal/Filters/Map/MapFilterViewController.swift
```diff
@@ -2,84 +2,87 @@
...
+        buttonItem.setTitleTextAttributes([.font: UIFont.bodyStrong], for: .normal)
+        return buttonItem
     }()
 
-    private lazy var searchLocationViewController: SearchLocationViewController = {
-        let searchLocationViewController = SearchLocationViewController()
-        searchLocationViewController.delegate = self
-        return searchLocationViewController
+    private lazy var mapContainerView: UIView = {
+        let view = UIView(frame: CGRect(x: 0, y: 0, width: 200, height: 200))
```
ninarg, line 43: When I don't init this with a height, I get a lot of breaking layout constraint issues when opening map search, because the map has size of zero when setting up the views. Is it okay to do this, or is there another solution?

osanoj: I think it is OK, I do it myself to avoid these issues. Very hard to work with 0 size views 😄

PR #336, opened 2020-05-14
Callout overlay dark mode fix

ninarg:
# Why?

A little dark mode bug sneaked into the polygon search. 

# What?

Callouts had a white background color. Now it has `bgPrimary` if dark mode, just like the bottom sheet.

# Show me

| Before | After |
| --- | --- |
| Image from iOS (1) | IMG_1807 |

### DIFF
--- a/Sources/Charcoal/Filters/Root/Views/CalloutOverlay.swift
+++ b/Sources/Charcoal/Filters/Root/Views/CalloutOverlay.swift
@@ -37,7 +37,7 @@ final class CalloutOverlay: UIView {
         let view = UIView(withAutoLayout: true)
         if #available(iOS 13, *),
             traitCollection.userInterfaceStyle == .dark {
-            view.backgroundColor = UIColor.bgSecondary.withAlphaComponent(0.8)
+            view.backgroundColor = UIColor.bgSecondary.withAlphaComponent(0.6)
         } else {
             view.backgroundColor = UIColor(white: 1, alpha: 0.8)
         }
--- a/Sources/Charcoal/Filters/Root/Views/CalloutOverlay.swift
+++ b/Sources/Charcoal/Filters/Root/Views/CalloutOverlay.swift
@@ -35,12 +35,7 @@ final class CalloutOverlay: UIView {
 
     private lazy var bodyView: UIView = {
         let view = UIView(withAutoLayout: true)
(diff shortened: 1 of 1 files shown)


Sources/Charcoal/Filters/Root/Views/CalloutOverlay.swift
```diff
@@ -35,7 +35,12 @@ final class CalloutOverlay: UIView {
 
     private lazy var bodyView: UIView = {
         let view = UIView(withAutoLayout: true)
-        view.backgroundColor = UIColor(white: 1, alpha: 0.8)
+        if #available(iOS 13, *),
+            traitCollection.userInterfaceStyle == .dark {
+            view.backgroundColor = UIColor.bgSecondary.withAlphaComponent(0.6)
+        } else {
+            view.backgroundColor = UIColor(white: 1, alpha: 0.8)
+        }
```
fespinoza, line 43: we do import FinniversKit from charcoal.
Then you can define a private extension on UIColor to define this background and use `dynamicColorIfAvailable`

ninarg: Did that now 👍 Still needed the dark mode check, since they had different alpha components. (Thought it should be more transparent for dark mode, since the button was barely visible through the overlay).

Sources/Charcoal/Filters/Root/Views/CalloutOverlay.swift
```diff
@@ -120,3 +125,9 @@ final class CalloutOverlay: UIView {
         delegate?.calloutOverlayDidTapInside(self)
     }
 }
+
+private extension UIColor {
+    class var overlayBackground: UIColor {
+        dynamicColorIfAvailable(defaultColor: .milk, darkModeColor: .darkIce)
```
fespinoza, line 131: I guess you can define it like

```
dynamicColorIfAvailable(defaultColor: .milk.withAlphaComponent(0.8), darkModeColor: .darkIce.withAlphaComponent(0.6))
```

ninarg: Aha, it actually works when I remove the `.` . Didn't know. 😛 Updated now.

PR #340, opened 2020-05-28
Feature: Filter-as-you-type for filtering subfilters

bstien:
# Why?
We want to implement a similar feature to what Android implemented in one of their hackdays. Some lists contains a lot items, like car/boat brand, and instead of scrolling endlessly the user should be able to narrow down the list with a query.

**Functionality**:
- The searchbar will be visible for filters with a subfilter count higher than what's defined through the constructor argument `searchbarSubfilterThreshold`.
- The keyboard will dismiss on drag within the tableview.
- Subfilters matching the query, or all if searchbar is empty, are copied into `scopedSubfilters`.
  - Subfilters prefixed with the query are prioritized to the top of the list.
- The list `scopedSubfilters` is used to populate the tableView.

# What?
- Add searchbar to top of view, if subfilters exceed the threshold.
- Pull out private class `FreeTextFilterSearchBar` into separate file and use this one.
  - It already has styling we need configured.
- Pull constraint for `topShadowView.bottomAnchor` into separate variable.

# Show me
Kapture 2020-05-28 at 14 01 54

### DIFF
--- a/Sources/Charcoal/Filters/FreeText/FreeTextFilterSearchBar.swift
+++ b/Sources/Charcoal/Filters/FreeText/FreeTextFilterSearchBar.swift
@@ -1,0 +1,34 @@
+import FinniversKit
+
+class FreeTextFilterSearchBar: UISearchBar {
+    // Makes sure to setup appearance proxy one time and one time only
+    private static let setupSearchQuerySearchBarAppereanceOnce: () = {
+        let textFieldAppearanceInRoot = UITextField.appearance(whenContainedInInstancesOf: [AppearanceColoredTableView.self])
+        textFieldAppearanceInRoot.adjustsFontForContentSizeCategory = true
+        textFieldAppearanceInRoot.defaultTextAttributes = [
+            NSAttributedString.Key.foregroundColor: UIColor.textAction,
+            NSAttributedString.Key.font: UIFont.bodyRegular,
+        ]
+
+        let textFieldAppearanceInSearch = UITextField.appearance(whenContainedInInstancesOf: [FreeTextFilterSearchBar.self])
+        textFieldAppearanceInRoot.adjustsFontForContentSizeCategory = true
+        textFieldAppearanceInSearch.defaultTextAttributes = [
+            NSAttributedString.Key.foregroundColor: UIColor.textPrimary,
+            NSAttributedString.Key.font: UIFont.bodyRegular,
+        ]
+
+        let barButtondAppearance = UIBarButtonItem.appearance(whenContainedInInstancesOf: [FreeTextFilterSearchBar.self])
+        barButtondAppearance.setTitleTextAttributes([.font: UIFont.bodyRegular])
+        barButtondAppearance.title = "cancel".localized()
+    }()
+
+    override init(frame: CGRect) {
+        _ = FreeTextFilterSearchBar.setupSearchQuerySearchBarAppereanceOnce
+        super.init(frame: frame)
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        _ = FreeTextFilterSearchBar.setupSearchQuerySearchBarAppereanceOnce
+        super.init(coder: aDecoder)
+    }
+}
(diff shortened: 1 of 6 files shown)


Sources/Charcoal/Filters/FreeText/FreeTextFilterSearchBar.swift
```diff
@@ -0,0 +1,38 @@
...
+        textFieldAppearanceInRoot.defaultTextAttributes = [
+            NSAttributedString.Key.foregroundColor: UIColor.textAction,
+            NSAttributedString.Key.font: UIFont.bodyRegular,
+        ]
+
+        let textFieldAppearanceInSearch = UITextField.appearance(whenContainedInInstancesOf: [FreeTextFilterSearchBar.self])
+        textFieldAppearanceInRoot.adjustsFontForContentSizeCategory = true
+        textFieldAppearanceInSearch.defaultTextAttributes = [
+            NSAttributedString.Key.foregroundColor: UIColor.textPrimary,
+            NSAttributedString.Key.font: UIFont.bodyRegular,
```
fespinoza, line 21: nitpick: we can remove the ` NSAttributedString.Key`, just `.font`

Sources/Charcoal/Filters/List/ListFilterViewController.swift
```diff
@@ -23,16 +23,32 @@ public final class ListFilterViewController: FilterViewController {
...
+        searchBar.searchBarStyle = .minimal
+        searchBar.delegate = self
+        searchBar.backgroundColor = Theme.mainBackground
+        searchBar.placeholder = "filterAsYouType.placeholder".localized()
+        searchBar.translatesAutoresizingMaskIntoConstraints = false
+        return searchBar
+    }()
+
     private let filter: Filter
+    private var filteredSubfilters: [Filter]
```
fespinoza, line 37: mmm maybe `scopedSubfilters`

Sources/Charcoal/Filters/List/ListFilterViewController.swift
```diff
@@ -63,12 +86,57 @@ public final class ListFilterViewController: FilterViewController {
...
+    }
+
+    // MARK: - Actions
+
+    @objc private func adjustForKeyboard(notification: Notification) {
+        guard let keyboardValue = notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue else { return }
+
+        var keyboardHeight = view.convert(keyboardValue.cgRectValue, from: view.window).height
+
+        if #available(iOS 11.0, *) {
```
fespinoza, line 129: uh, do we support less than iOS 11 in charcoal? I guess we shouldn't as we don't test it

bstien: No, I think you're right. Copied this code from another place, so I'll change there as well

Sources/Charcoal/Filters/List/ListFilterViewController.swift
```diff
@@ -159,3 +224,18 @@ extension ListFilterViewController: UITableViewDelegate {
...
+// MARK: - UISearchBarDelegate
+
+extension ListFilterViewController: UISearchBarDelegate {
+    public func searchBar(_ searchBar: UISearchBar, textDidChange searchText: String) {
+        if searchText.isEmpty {
+            scopedSubfilters = filter.subfilters
+        } else {
+            let searchTextLowercased = searchText.lowercased()
+            scopedSubfilters = filter.subfilters.filter { $0.title.lowercased().contains(searchTextLowercased) }
+        }
```
hakloev, line 237: Just a little tip or "nice-to-have" thing we discovered when adding a similar feature in `ad-insertion-ios`:

It could be a good idea to sort the matches by prefix matching here as well, to avoid "less" important matches occurring before the exact matches. I.e. if you search for "Land Rover" and type "La" you would expect "Land Rover" and "Lamborghini" to occur before "Cadillac" 🙃

```
scopedSubfilters = filter.subfilters
    .filter { $0.title.lowercased().contains(searchTextLowercased) }
    .sorted {
        let first = $0.title.lowercased().hasPrefix(query) ? 0 : 1
        let second = $1.title.lowercased().hasPrefix(query) ? 0 : 1
        return first < second
    }
```

bstien: That sounds pretty clever! I'll try it out 🌟

bstien: I think this makes much more sense 😍
Screenshot 2020-05-28 at 14 59 06

hakloev: Nice! :D I think so too :)

PR #344, opened 2020-07-15
Support search filter tags

ninarg:
# Why?

We want to display search filter tags on top of search results in the app. Currently, our search filters are encapsulated in charcoal, so we need a way to get the relevant info out of here.

# What?

- Added public var **selectedFilters**, which are filters displayed as search filter tags.
- Added public method to **remove a filter**, used when tapping delete button in search filter tag
- Added public methods for a filter’s title and validity, which is needed as part of view model for search filter tags
- Added public method to shortcut to a filter. When the label of a search filter tag is pressed, `shortcutToFilter` is called, to open the bottom sheet with that specific filter already showing
    - In order to do this, I also added methods to figure out all parents of this filter, and push their view controllers too, so you always get the correct view hierarchy when shortcutting to a filter.
    - When shortcutting to free text filter, we make sure that the search bar is first responder, so the keyboard opens. 
    - When shortcutting to an inline filter, we scroll to that position in the inline filter scrollview.  
- Had to do some fixes in layout constraints for `GridFilterViewController` and `ListFilterViewController`. When implementing shortcuts to filters, I discovered layout issues with the bottom anchor constraint for the first filter that was pushed on top of `rootFilterViewController`. The view would not be visible, even though it should be aligned to the top of the bottomButton. The weird thing is that everything works fine if the view hiearchy is like this: `rootFillterVC -> filterVC -> filterVC`, but for `rootFilterVC -> filterVC`, the filterVC is just not visible.. I aligned it to the bottom of the view instead, and added more space equal to the height of the button. There probably exists a better solution than this, but nothing else seemed to work for me.. 🤔


ninarg:
Ready for review! 😄

Sources/Charcoal/Filters/List/ListFilterViewController.swift
```diff
@@ -78,6 +83,10 @@ public final class ListFilterViewController: FilterViewController {
 
     public override func showBottomButton(_ show: Bool, animated: Bool) {
         super.showBottomButton(show, animated: animated)
+        if show {
+            tableViewBottomAnchorConstraint.constant = -bottomButton.frame.height
```
ninarg, line 87: @finn-no/ios This is the last PR that needs a review before search filter tags are done! I would especially like feedback on my layout fixes in this file. I'm not 100% sure if it's safe, but didn't find another way to solve it. Read the PR description for more info 😄

Sources/Charcoal/Filters/Stepper/StepperViewController.swift
```diff
@@ -51,8 +51,8 @@ public final class StepperFilterViewController: FilterViewController {
         ])
     }
 
-    public override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
+    public override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
         topConstraint.constant = (view.frame.height - bottomButton.intrinsicContentSize.height) / 2
```
osanoj, line 56: Not sure why this is moved to `viewDidLayoutSubviews`, but I assume it fixes some issue 😄

osanoj (approved):
Code looks good, can have a run through on the UI when it is merged and builds with Xcode 12 😄

Sources/Charcoal/Filters/Stepper/StepperViewController.swift
ninarg: Yes, there was a spacing issue. 😄  I don't think it's safe to use the frames before `viewDidLayoutSubviews`, because they're not always correct.

Sources/Charcoal/CharcoalViewController.swift
```diff
@@ -96,6 +100,44 @@ public final class CharcoalViewController: UINavigationController {
...
     }
 
+    public func removeFilter(_ filter: Filter) {
+        selectionStore.removeValues(for: filter)
+        handleFilterSelectionChange(from: .externalSearchFilterTag)
+    }
+
+    public func title(for filter: Filter) -> SelectionTitle? {
+        let titles = selectionStore.titles(for: filter)
+        guard titles.count == 1 else { return nil }
```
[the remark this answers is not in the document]
ninarg: I added this guard just in case the filter had more than one title. Filters with subfilters have multiple titles. Thought it might be confusing if you only got the first title back, if the filter happened to have multiple titles. 🤔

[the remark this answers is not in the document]
osanoj: Ah yes they could have multiple titles, that makes your code and my suggestion different 👍 😄

PR #347, opened 2020-08-28
Make map filter view controller public

vadymmarkov:
# Why?

In order to use the map filter without `CharcoalViewController`.

# What?

Make map filter view controller public.

# Show me

No UI changes.


fespinoza (approved):
LGTM

PR #348, opened 2020-09-02
Filter for knots (boat)

bstien:
# Why?
There's been added a speed filter for the boat market. The unit for this filter should be knop/knots.

# What?
- Added the filter and localized strings.

# Show me
Screenshot 2020-09-02 at 15 05 28


osanoj (approved):
🚤

PR #350, opened 2020-09-18
Remove outdated callouts

ninarg:
# Why?

We have some old callouts that we don't need anymore.

# What?

- Delete region reform callout
- Delete polygon search callout
- Update tests


bstien (approved):
Nice 😎🗑️

PR #361, opened 2021-01-05
Add release script

osanoj:
# Why?
We already have fastlane in this project, so why not copy the release part from FinniversKit.

# What?
Should now be very similar to when making FinniversKit releases. See updated README.md for release instructions 😄


ninarg (approved):
awesome 😸

ag12 (approved):
Have no idea what you have done. But I will approve it 😄

PR #364, opened 2021-02-10
Feature: SPM support

bstien:
# Why?
We want to introduce SPM support to this library.

# What?
- Added `Package.swift`.
- Moved `PolygonData` from FINNSetup to Charcoal
  - FINNSetup should be dependent on Charcoal, not the other way around.
- Remove FINNSetup. #366 
  - No longer needed, since we have FilterTranslator.
- Rewrite demo to show dummy-filters. #365 
- Add snapshot tests – like in NC. #367
- Remove Carthage. #367

# Show me
_No UI changes_


bstien:
I'll create PRs pointing to this branch for the stuff mentioned in the todo list.

Package.resolved
```diff
@@ -0,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "FinniversKit",
+        "repositoryURL": "https://github.com/finn-no/FinniversKit.git",
+        "state": {
+          "branch": null,
+          "revision": "0235727d2c41a228e712ad3a2babc62a4430bd8d",
+          "version": "76.0.0"
```
ninarg, line 10: this version number is outdated - run `swift package resolve`? 🤔

ninarg (approved):
Let's merge 💣

ninarg:
We should also remove the lines about carthage support in readme

PR #365, opened 2021-02-11
Demo: Remove FINNSetup

bstien:
# Why?
Since we'll be moving the filter parsing and translation away from Charcoal/FINNSetup we should use the demo in this repo to verify the filters' layout and design, rather than verifying setup from a JSON response.

This PR points to my `spm-support` branch in #364.

# What?
- Refactored demo to show off our filters in different scenarios.
- Demo is no longer dependent on `FINNSetup`.
- Removed JSON files used for setting up the demo.

# Show me

https://user-images.githubusercontent.com/1901556/107627052-ff3f5b80-6c5e-11eb-84d9-d37c834a796b.mp4


hakloev (approved):
LGTM

PR #367, opened 2021-02-15
SPM support: Snapshot tests and remove Carthage/AppCenter

bstien:
_Note: I checked out this branch from #365 before it was merged, so the commits made towards that branch seems to be listed in this branch as well._

# Why?
The demo views added in #365 should have snapshot tests, plus we don't need Carthage nor an AppCenter demo anymore.

# What?
I've done the same setup as I've done with FilterTranslator: creating a new project where the Swift package has been drag-and-dropped into, adding a new test target and placing all tests (including the package tests) as part of the `CharcoalDemo` test scheme.

Both Carthage- and AppCenter-support have been removed.

## Snapshot tests
I had some issues trying to get consistently successful snapshot tests in the beginning, and the reason was because Xcode and AppCode produced different images on the same test device. So in order to prevent this from happening on the CI I've configured to capture the snapshots on an iPhone 8. Hopefully that should work.

Initially I also had a test added for the `MapFilterDemoViewController`, but because the map loading is async I wasn't able to snapshot it consistently. So I removed it for now and I'm hoping we can find another way to test this one.


bstien:
This one is now building on CircleCI and ready for review

bstien:
I've removed the workspace here as well in favor of the plain Xcode project

ninarg:
Awesome work! 🥇  
Great with some snapshot tests. I noticed the snapshot for range filter on iPad light mode (`testRangeFilter.iPad_lightMode.png`) is a bit off though, while dark mode is correct. That's a bit weird. Do you have any idea why that is? 🤔

bstien:
> Awesome work! 🥇
> Great with some snapshot tests. I noticed the snapshot for range filter on iPad light mode (`testRangeFilter.iPad_lightMode.png`) is a bit off though, while dark mode is correct. That's a bit weird. Do you have any idea why that is? 🤔

I have no idea. I also see that the same view/test have minor differences between light- and dark-mode. Like the font for the searchBar in `testMultipleVerticals.iPad_xmode`.

We take light-mode snapshots first and then dark-mode, so it could be that the view isn't fully set up until dark-mode comes around.

I'll see if I can find any answers within the `SnapshotTesting` repo :)

Sources/Charcoal/Filters/Range/RangeFilterView.swift
```diff
@@ -97,6 +97,7 @@ final class RangeFilterView: UIView {
     }
 
     override func layoutSubviews() {
+        sliderInputView.layoutIfNeeded()
```
bstien, line 100: I'm not sure why `super.layoutSubviews()` isn't called from here, but calling that messes up the snapshot test for iPhone 🤷‍♂️

ninarg: Weird 🤔

ninarg (approved):
Range filter snapshot looks correct now 👍

PR #369, opened 2021-09-20
Remove warnings and delete outdated code

ninarg:
# Why?

Always nice to clean up 🧽 

# What?

Remove warnings
- Rename localizable file to `nb` instead of `en`, since it's Norwegian.

Delete outdated stuff
- Changelog related to demo
- Info about creating demo release - we don't distribute a demo app anymore

Also fixed the import info in readme, current didn't work.

# Show me

No UI change.


bstien (approved):
🥳🥳

PR #376, opened 2022-02-25
Fix: Memory leak for filters

ninarg:
# Why?

I found a memory leak.

# What?

Each filter keeps a array of sub filters. The sub filters also keep a reference to their parent filter, which should be weak to avoid memory leak.


hakloev (approved):
Great find! 🎉

PR #377, opened 2022-04-25
Remove callout to næringstorget

ninarg:
# Why?

We used to promote næringstorget with a callout, but the project has been shut down.

# What?

- Delete the callout

# Show me

Imagine there's no callout when opening the motor market filters.


robinsalehjan (approved):
🚢

PR #380, opened 2022-06-28
Bump FinniversKit to version 93.7.0

ninarg:
# Why?

Bumping FinniversKit to latest version, so it's possible to bump it in packages that depend on Charcoal. 

# What?

Bum FinniversKit to version 93.7.0

A breaking change in FinniversKit is that iOS 13 is the minimum target, so we can remove `userInterfaceStyleSupport` config. Also removed a iOS 13 check for this package.

# Show me

No UI.


ninarg:
Closing this one and opening a new one on another branch - since I wasn't able to delete cache of failing tests.

PR #383, opened 2022-10-26
Fix truncated units on iOS 16

osanoj:
# Why?
The units text was truncated on iOS 16

# What?
Apparently the `tailIndent` of `NSMutableParagraphStyle` works differently on iOS 16, but there is no change on previous iOS versions when it is removed so that is good 😄 

Also started updated the project quite a bit, but then realized it made this fix a bit hard to find, so there will be another PR that fixes all that.

# Show me
| iOS 16 Before | iOS 16 After |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2022-10-26 at 15 13 06 | Simulator Screen Shot - iPhone 8 - 2022-10-26 at 15 14 35 |

| iOS 15.4 Before | iOS 15.4 After |
| --- | --- |
| Simulator Screen Shot - iPhone 8 - 2022-10-26 at 15 12 38 | Simulator Screen Shot - iPhone 8 - 2022-10-26 at 15 15 48 |

### DIFF
--- a/Sources/Charcoal/Filters/Range/Input/NumberInputView.swift
+++ b/Sources/Charcoal/Filters/Range/Input/NumberInputView.swift
@@ -193,7 +193,6 @@ final class NumberInputView: UIView {
         let style = NSMutableParagraphStyle()
         style.alignment = .justified
         style.headIndent = .spacingS
-        style.tailIndent = -.spacingS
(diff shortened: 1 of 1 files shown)


bstien (approved):
Nice find!

AndreyMomot (approved):

PR #384, opened 2022-10-26
Update "all the things"

osanoj:
# Why?
A lot was outdated in this project.

# What?
- Setting minimum version to iOS 14
- Removed outdated available-checks
- Updated gems
- Updated snapshot lib and configured perceptualPrecision
- Fix SwiftLint config and remove all the force unwraps in the code
- Remove swiftformat that was not actually working

# Show me
Should not be visible 😄


Sources/Charcoal/Extensions/UIKit/UIViewExtensions.swift
```diff
@@ -6,11 +6,7 @@ import UIKit
 
 extension UIView {
     var windowSafeAreaInsets: UIEdgeInsets {
-        if #available(iOS 11.0, *) {
```
ninarg, line 9: 11 😮

Sources/Charcoal/Filters/Root/RootFilterViewController.swift
```diff
@@ -474,44 +469,45 @@ private extension RootFilterViewController {
...
-        }
+            let freeTextFilterViewController = FreeTextFilterViewController(filter: freeTextFilter, selectionStore: selectionStore)
+            self.freeTextFilterViewController = freeTextFilterViewController
+            freeTextFilterViewController.delegate = self
+            freeTextFilterViewController.filterDelegate = freeTextFilterDelegate
+            freeTextFilterViewController.filterDataSource = freeTextFilterDataSource
 
-        let searchBar = freeTextFilterViewController!.searchBar
-        tableView.addSubview(searchBar)
+            tableView.addSubview(freeTextFilterViewController.searchBar)
```
ninarg, line 478: Was it an intentional change to add `searchBar` as subview only on 1st call of this function? It is `.removeFromSuperview()` up there, so might cause trouble if this func is supposed to add it back? 🤔

PR #388, opened 2023-01-03
Upgrades: CircleCI uses Xcode 14.1 with iOS 16.1 - upgrade FinniversKit to version 101.0.0

ninarg:
Need to upgrade FinniversKit to version 101.0.0.

In the process, tests were suddently failing, so I decided to upgrade CircleCI to use Xcode 14.1 and iOS 16.1. That solved the issue. Decided to set device to iPhone 14 Pro, since that's consistent with other packages. 
Rerecorded all snapshots.


AtishPapa (approved):
LGTM 👍

PR #392, opened 2023-09-05
English translation

ninarg:
# Why?

Charcoal should be available in English for debug.

# What?

Translate it.

### DIFF
--- a/UnitTests/Charcoal/Selection/FilterSelectionStoreTests.swift
+++ b/UnitTests/Charcoal/Selection/FilterSelectionStoreTests.swift
@@ -209,7 +209,7 @@ final class FilterSelectionStoreTests: XCTestCase {
 
         store.removeValues(for: lowValueFilter)
         store.setValue(100, for: highValueFilter)
-        XCTAssertFalse(store.titles(for: filter).first!.value.isEmpty)
+        XCTAssertTrue((store.titles(for: filter).count == 1))
         XCTAssertFalse(store.titles(for: filter).first!.value.isEmpty)
         XCTAssertFalse(store.titles(for: filter).first!.accessibilityLabel.isEmpty)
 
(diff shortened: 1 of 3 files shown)


ag12 (approved):
horsepowers

UnitTests/Charcoal/Selection/FilterSelectionStoreTests.swift
```diff
@@ -200,32 +200,22 @@ final class FilterSelectionStoreTests: XCTestCase {
...
-        XCTAssertFalse(store.titles(for: filter)[0].value.isEmpty)
-        XCTAssertFalse(store.titles(for: filter)[0].accessibilityLabel.isEmpty)
-
-//        XCTAssertEqual(store.titles(for: filter), [
-//            SelectionTitle(value: "10 - 100 kr", accessibilityLabel: "Fra 10 Opptil 100 kroner"),
-//        ])
+        XCTAssertTrue(store.titles(for: filter)[0].value.contains("10"))
+        XCTAssertTrue(store.titles(for: filter)[0].value.contains("100"))
+        XCTAssertTrue(store.titles(for: filter)[0].accessibilityLabel.contains("10"))
+        XCTAssertTrue(store.titles(for: filter)[0].accessibilityLabel.contains("100"))
```
bstien, line 218: This gives a false sense of security, since `100` will always match `10`.
Can we either use regex or maybe break up the string, i.e. by spaces, and then check if the array contains the values we expect?

bstien (approved):

UnitTests/Charcoal/Selection/FilterSelectionStoreTests.swift
ninarg: Veeery good point, fixed ✅

PR #395, opened 2023-09-25
Xcode 15 compat and iOS 15 minimum

BjornRuud:
# Why?

Bump min iOS to 15 and fix Xcode 15 compatibility.

# What?

- Bump versions.
- Disable asset generation.
- Update snapshots.
- Set Xcode 15 as CI target.


robinsalehjan (approved):
🥇 🥇

BjornRuud:
Snapshot updates was a bit premature. We're going to stay on iOS 16 for snapshots for now. I'll revert that part.

PR #399, opened 2023-10-18
Support multiple currencies

ninarg:
# Why?

We need to support euros as well as NOKs. 

# What?

Make an API for injecting local currencies into Charcoal. For the currency filter, we localise using local currencies instead of current hard coding to "kr" or "€". 

Attach a `unit` String to currency filter (will help us fallback on backend's unit whenever we don't have a local currency configured).

Added a new test covering this change.

# Show me

We were already supporting euros before this change due to fallback on backend's unit which is also `€`, but this is how it looks:

### DIFF
--- a/Sources/Charcoal/Resources/en.lproj/Localizable.strings
+++ b/Sources/Charcoal/Resources/en.lproj/Localizable.strings
@@ -34,7 +34,6 @@
 
 "unit.centimeters.value" = "cm";
 "unit.cubicCentimeters.value" = "cm3";
-"unit.currency.value" = "kr";
 "unit.feet.value" = "feet";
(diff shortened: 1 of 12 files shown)


Sources/Charcoal/Resources/en.lproj/Localizable.strings
```diff
@@ -47,7 +46,6 @@
 
 "unit.centimeters.accessibilityValue" = "centimetre";
 "unit.cubicCentimeters.accessibilityValue" = "cubic centimetre";
-"unit.currency.accessibilityValue" = "crowns";
```
bstien, line 50: 👑👑👑

bstien (approved):
LGTM!

PR #401, opened 2024-01-11
Update semantic colors to prepare for new Tori colors

ninarg:
# Why?

There's been a change in FinniversKit regarding semantic colors. Now `btnPrimary` is supposed to be used only on CTA/primary buttons. This is to prepare for new Tori colors where `btnPrimary` will stand out with a watermelon color, while the rest of the components will be blueberry.

# What?

Replace `btnPrimary` with other semantic colors based on the context they're used in.  This change is not visible in FINN, since all of the replacement colors are blue (`nmpBrandControlSelected`, `nmpBrandDecoration`, `textAction`).

Bump FinniversKit to new version. The blue color changed in FINN dark mode. Rerecorded snapshots. Some of the light mode snapshots were also changed, seems like the position of things were adjusted like 1 pixel, hope they still pass through the tests.

# Show me

Blue color changed in dark mode, see snapshot diffs.


BjornRuud:
Why was `testRangeFilter` snapshots updated for light mode?

BjornRuud (approved):

ninarg:
> Why was `testRangeFilter` snapshots updated for light mode?

The difference I could see was a shift in 1 pixel on the position of items. Not sure why it happened, so will see if it passes through CircleCI tests.

PR #407, opened 2024-04-17
BTPB-2280: Add a new kind to Filter model called freeTextOnly 

nssina:
# Why?

We needed to add a new filter kind to handle a new state for the search bar in the search results view in FINN and Tori.

# What?

- Added new `freeTextOnly` option to `Filter.Kind` model
- Handle FreeTextFilterViewController presentation based on filter model kind

### DIFF
--- a/Sources/Charcoal/Filters/FreeText/FreeTextFilterViewController.swift
+++ b/Sources/Charcoal/Filters/FreeText/FreeTextFilterViewController.swift
@@ -31,6 +31,10 @@ public class FreeTextFilterViewController: ScrollViewController {
 
     weak var delegate: FreeTextFilterViewControllerDelegate?
 
+    // MARK: - Internal Properties
+    
+    var isFreeTextOnlySelected = false
+
     // MARK: - Private Properties
 
     private var didClearText = false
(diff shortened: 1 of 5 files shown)


Sources/Charcoal/Filters/FreeText/FreeTextFilterViewController.swift
```diff
@@ -223,13 +228,22 @@ extension FreeTextFilterViewController: UISearchBarDelegate {
 
 private extension FreeTextFilterViewController {
     func returnToSuperView() {
+        if isFreeTextOnlySelected {
+            dismissView()
```
ninarg, line 232: Can we move `isFreeTextOnlySelected = false` here? `dismissView()` could potentially be used in another context than for "free text only", but then it shouldn't necessarily reset this unrelated var.

Sources/Charcoal/Filters/Root/RootFilterViewController.swift
```diff
@@ -28,6 +28,7 @@ final class RootFilterViewController: FilterViewController {
         didSet { freeTextFilterViewController?.filterDataSource = freeTextFilterDataSource }
     }
 
+    var isFreeTextOnlySelected: Bool = false
```
ninarg, line 31: I think the naming of this var can be confusing: `isFreeTextOnlySelected`.
It sounds like setting it will enable the free text filter only. But that only happens in case you also shortcut to this filter at the same time, not in other contexts. If that makes sense?
What setting this var really does, is to make `FreeTextFilterViewController` dismiss filters on the next selection.

Perhaps we can rename it to something like that?
`dismissFiltersOnNextFreeTextSelection`? Or other, if you have a better name 😄

ninarg (approved):
LGTM! 😄 Added a couple suggestions

Sources/Charcoal/Filters/Root/RootFilterViewController.swift
nssina: Agree, I'll update the name with the one you suggest :D

PR #408, opened 2024-07-12
Verticals: Give flexibility by adding separate configure functions and reload mechanism

ninarg:
# Why?

Charcoal was built under the assumption verticals are available when initialising with filters. That has now changed, verticals will be loaded separately - and we need a mechanism for handling it when it fails.

# What?

- Add a public function to configure verticals - `public func configure(with verticals: [Vertical]) `
- Add a public function to show and hide reload verticals button - `public func updateReloadVerticalsButton(isVisible: Bool)`
- Add a delegate function for handling clicks on reload button

## Bonus cleanup
I noticed we have a bunch of map selection delegates that are unused in the app. So decided to delete them now that I will make a breaking change anyways.

# Show me

As you see in this demo, whenever verticals failed we display a button on the left hand side to fetch verticals.
After a successful fetch, the verticals are shown.

https://github.com/user-attachments/assets/70e94ab7-3292-4de9-a943-90a81d453219


ag12 (approved):
👍

PR #409, opened 2024-07-29
Update color and spacing

milad-schibsted:
# Why?

We are going to start using WARP colors instead of old colors, to make it easier to move forward with other brands.

# What?
This is mostly find and replace of `spacing` and `colors` from the old system to WARP. 

# UI Changes

There are lots of UI changes, mainly color changes.


ag12 (approved):
legit

PR #414, opened 2024-08-20
CLLocationManager warnings fix

ninarg:
# Why?

We have a runtime warning when using CLLocationManager's `locationServicesEnabled()`

Screenshot 2024-08-20 at 08 43 17

# What?

According to [Apple's documentation](https://developer.apple.com/documentation/corelocation/cllocationmanager/authorizationstatus-swift.property), we should use a delegate function [locationManagerDidChangeAuthorization(_:)](https://developer.apple.com/documentation/corelocation/cllocationmanagerdelegate/locationmanagerdidchangeauthorization(_:)) to check the app's initial authorization state. The system guarantees to call it after init of `CLLocationManager`.

Added a new `LocationService` to share logic between polygon and radius filter. The users of it can now get the authorization status, but the difference from before is that it's now async, since we have to wait for the delegate function to be called first.

# Show me

No UI change.

### DIFF
--- a/Sources/Charcoal/Filters/Map/LocationService.swift
+++ b/Sources/Charcoal/Filters/Map/LocationService.swift
@@ -43,7 +43,7 @@ extension CLAuthorizationStatus {
             return true
         case .notDetermined, .restricted, .denied:
             return false
-        default:
+        @unknown default:
             return false
         }
     }
--- a/Sources/Charcoal/Filters/Map/LocationService.swift
+++ b/Sources/Charcoal/Filters/Map/LocationService.swift
@@ -1,0 +1,50 @@
+import Foundation
+import CoreLocation
+
+class LocationService: NSObject {
+    private let locationManager = CLLocationManager()
+    private var authorizationContinuations = [CheckedContinuation<CLAuthorizationStatus, Never>]()
+    private var didSyncAuthStatus = false
+
+    override init() {
+        super.init()
+        locationManager.delegate = self
+    }
+
+    func authorizationStatus() async -> CLAuthorizationStatus {
+        if didSyncAuthStatus {
+            return locationManager.authorizationStatus
+        }
+        return await withCheckedContinuation { continuation in
+            authorizationContinuations.append(continuation)
+        }
+    }
+
+    func requestWhenInUseAuthorization() {
+        locationManager.requestWhenInUseAuthorization()
+    }
+}
+
+extension LocationService: CLLocationManagerDelegate {
+    func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
+        didSyncAuthStatus = true
+
+        for continuation in authorizationContinuations {
+            continuation.resume(returning: manager.authorizationStatus)
+        }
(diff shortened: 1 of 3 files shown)


Sources/Charcoal/Filters/Map/LocationService.swift
```diff
@@ -0,0 +1,50 @@
...
+
+extension CLAuthorizationStatus {
+    var isLocationAuthorized: Bool {
+        switch self {
+        case .authorizedAlways, .authorizedWhenInUse:
+            return true
+        case .notDetermined, .restricted, .denied:
+            return false
+        default:
+            return false
```
Breiby, line 47: Might be wise to add unknown here to make sure we handle possible future cases.

```suggestion
        @unknown default:
            return false
```

Sources/Charcoal/Filters/Map/LocationService.swift
```diff
@@ -0,0 +1,50 @@
...
+        locationManager.delegate = self
+    }
+
+    func authorizationStatus() async -> CLAuthorizationStatus {
+        if didSyncAuthStatus {
+            return locationManager.authorizationStatus
+        }
+        return await withCheckedContinuation { continuation in
+            authorizationContinuations.append(continuation)
+        }
```
Breiby, line 20: Should we weakify `self` here?

Breiby (approved):
LGTM! 🙌

Sources/Charcoal/Filters/Map/LocationService.swift
ninarg: Apparently it's not needed. Here's what ChatGPT says:

In this particular case, you do not need to weakly capture self inside the withCheckedContinuation block. Here’s why:

**1. Continuation Blocks and Memory Management:**
- The withCheckedContinuation block will be executed immediately and is not stored or retained beyond the duration of the call itself.
- Since the continuation is handled immediately and there’s no possibility of a strong reference cycle being created (where self would keep the block alive and the block would keep self alive), there's no need to weakly capture self in this context.

**2. Scope and Lifetime:**
- The self reference in the block is used to append the continuation to the authorizationContinuations array. Since the block is not stored and is only used in the context of this method, it won’t outlive the scope of the authorizationStatus method call.

**3. Best Practices:**
- Weakly capturing self is necessary when the block is retained by self or by an object that could cause a retain cycle, but that’s not the case here. The continuation is a temporary construct used to manage the async/await behavior and doesn’t introduce a strong reference cycle.

Conclusion:
In this scenario, you can safely capture self strongly within the withCheckedContinuation block. There’s no risk of a retain cycle, and the continuation is not retained beyond the method scope.

If the withCheckedContinuation were capturing self for an asynchronous operation that might last longer than the method scope (for example, if you were passing a closure to another long-lived object that could retain it), then you would need to consider capturing self weakly. But in this specific case, it’s not necessary.

Sources/Charcoal/Filters/Map/LocationService.swift
ninarg: Smart, done ✅

PR #415, opened 2024-08-20
Improved location sharing alert with button to open location settings

ninarg:
# Why?

We currently display an alert with some vague outdated description on how to make it work.

# What?

Improved it by guiding the user, now we provide a button to open settings for the app, where the user easily can change their location sharing.

This alert is the same we use other places in the app.

# Show me

| Before | After |
| --- | --- |
| IMG_2566 | IMG_2569 |


ag12 (approved):
⚡

PR #417, opened 2024-09-04
Remove font provider

mohsen-schibsted:
# Why?

We are going to start using WARP Typography instead of old fonts, to make it easier to move forward with other brands.

# What?

Removing Finniverskit FontProvider so everywhere we are forced to use WARP Typography.

# Version Change

Since it is only changing the way we set the fonts it should be minor

# UI Changes

Changes in some fonts


ag12:
I think we need to upload new snapshots 👍

ag12:
cc: @mohsen-schibsted loks good to me! its seems the fonts has changed. really hard to notice 😅

PR #418, opened 2024-09-10
Remove font provider

mohsen-schibsted:
# Why?

We are going to start using WARP Typography instead of old fonts, to make it easier to move forward with other brands.

# What?

Removing Finniverskit FontProvider so everywhere we are forced to use WARP Typography.

# Version Change

Since it is only changing the way we set the fonts it should be minor

# UI Changes

Changes in some fonts


ag12 (approved):
🚀

ag12:
do you need med to run the tests @mohsen-schibsted ? :)

mohsen-schibsted:
> do you need me to run the tests @mohsen-schibsted ? :)

That would be much appreciated @ag12 , thank you :D

PR #420, opened 2024-11-19
configure default map filter mode

thorbjxrn:
# Why?

we would like to have polygon search as the default, so adding the possibility to configure the default map filter mode.

also see https://github.schibsted.io/finn/ios-app/pull/6200 for context

**PR closed due to charcoal moved into main repo**

# What?

added an enum to the initialiser and removed a function in setup that seemed redundant. 

# Show me

| Before | After |
| --- | --- |
| radius search | polygon search |

### DIFF
--- a/Sources/Charcoal/CharcoalViewController.swift
+++ b/Sources/Charcoal/CharcoalViewController.swift
@@ -32,6 +32,8 @@ public final class CharcoalViewController: UINavigationController {
         didSet { configure(with: filterContainer) }
     }
 
+    public var defaultMapMode: MapFilterMode = .radius
+
     public weak var textEditingDelegate: CharcoalViewControllerTextEditingDelegate?
     public weak var mapDelegate: CharcoalViewControllerMapDelegate?
     public weak var selectionDelegate: CharcoalViewControllerSelectionDelegate?
@@ -316,6 +318,7 @@ extension CharcoalViewController: FilterViewControllerDelegate {
                 locationNameFilter: locationNameFilter,
                 bboxFilter: bboxFilter,
                 polygonFilter: polygonFilter,
+                defaultMode: defaultMapMode,
                 selectionStore: selectionStore
             )
             mapFilterViewController.searchLocationDataSource = searchLocationDataSource
--- a/Sources/Charcoal/Filters/Map/MapFilterViewController.swift
+++ b/Sources/Charcoal/Filters/Map/MapFilterViewController.swift
@@ -32,6 +32,17 @@ public class MapFilterViewController: FilterViewController {
         }
     }
 
+    public var defaultMapMode: MapFilterMode {
+            didSet {
+                switch defaultMapMode {
+                    case .polygon:
+                        selectedViewController = mapPolygonFilterViewController ?? mapRadiusFilterViewController
+                    case .radius:
+                        selectedViewController = mapRadiusFilterViewController
+             }
+        }
+    }
+
     private lazy var toggleButton: UIBarButtonItem = {
         let buttonItem = UIBarButtonItem(
             title: nil,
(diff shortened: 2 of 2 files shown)


Sources/Charcoal/CharcoalViewController.swift
```diff
@@ -32,6 +32,8 @@ public final class CharcoalViewController: UINavigationController {
         didSet { configure(with: filterContainer) }
     }
 
+    public var defaultMapMode: MapFilterMode = .radius
```
robinsalehjan, line 35: I suggest that we turn this a computed property and override `didSet` to select the correct `selectedViewController` which makes it consistent with the rest of the interface; and will update the controller when the value changes as well.👍 

```suggestion
    public var defaultMapMode: MapFilterMode = {
        didSet {
            switch defaultMode {
                case .polygon:
                    selectedViewController = mapPolygonFilterViewController ?? mapRadiusFilterViewController
                case .radius:
                    selectedViewController = mapRadiusFilterViewController
         }
    }
}
```

Sources/Charcoal/Filters/Map/MapFilterViewController.swift
```diff
@@ -141,14 +152,6 @@ public class MapFilterViewController: FilterViewController {
 
         navigationItem.rightBarButtonItem = toggleButton
 
-        if let bboxFilter = bboxFilter,
-            let polygonFilter = polygonFilter,
-            selectionStore.isSelected(polygonFilter) || selectionStore.isSelected(bboxFilter) {
-            selectedViewController = mapPolygonFilterViewController
-        } else {
-            selectedViewController = mapRadiusFilterViewController
-        }
```
ninarg, line 150: We can't remove this code, it makes sure if you already have a filter selected, it will display that filter.
Now there's a problem that can be reproduced like this:
1. Open realestate search (it shows polygon map by default), but switch over to radius search
2. Make a search with radius filter
3. Close bottom sheet
4. Open bottom sheet, click map filter
5. Problem: Now the map filter displays polygon search, but radius search is the active filter

I think we can fix this by using the default mode if you have neither a polygon, bbox or radius filter selected (instead of using radius in that case).
Then we might be able to remove the selection of view controller from init as well, will make it simpler if it's only here.

thorbjxrn: i think we'll need it in the init also, as selectedViewController needs to be instantiated before function calls (self). or is there a good way to set it outside the init?

Sources/Charcoal/Filters/Map/MapFilterViewController.swift
```diff
@@ -72,8 +92,17 @@ public class MapFilterViewController: FilterViewController {
...
         }
 
+        defaultMapMode = defaultMode
+        switch defaultMapMode {
+        case .polygon:
+            selectedViewController = mapPolygonFilterViewController ?? mapRadiusFilterViewController
+        case .radius:
+            selectedViewController = mapRadiusFilterViewController
+        }
+
```
ninarg, line 102: hmm, this ended up being more complicated than it needs to be because of the computed property and needing to call the same function also later in the setup function. Right now we set the default view controller 3 times.
What if we change things a bit.
We can simply do one single thing in init: set the `selectedViewController` to whatever is `defaultMapMode`.
We can remove `defaultMapMode` variable in the class, there's no need to change it later.

Then in `setup`, we can simply change the default view controller to whatever filter is selected, but if no filter is selected, do nothing. If we do nothing there, the default view controller will still be selected. So we don't need an else there. (then we can also delete the `setMapViewToDefaultMode` function)

Does this make sense to you? 😄 
So we only need to have this logic in _one_ place where it is needed, and that is in init.

Sources/Charcoal/Filters/Map/MapFilterViewController.swift
```diff
@@ -46,9 +66,10 @@ public class MapFilterViewController: FilterViewController {
 
     private let bboxFilter: Filter?
     private let polygonFilter: Filter?
+    private let radiusFilter: Filter?
 
     public init(title: String, latitudeFilter: Filter, longitudeFilter: Filter, radiusFilter: Filter,
-                locationNameFilter: Filter, bboxFilter: Filter?, polygonFilter: Filter?, selectionStore: FilterSelectionStore) {
+                locationNameFilter: Filter, bboxFilter: Filter?, polygonFilter: Filter?, defaultMode: MapFilterMode = .radius, selectionStore: FilterSelectionStore) {
```
ninarg, line 72: Can we change the param name to `defaultMapMode` so it's easy to see what it means on caller side?

ninarg (commented):
This code has to be moved to the app repo now, since Charcoal has been moved there 😄

Sources/Charcoal/Filters/Map/MapFilterViewController.swift
thorbjxrn: I'll try it out! Thanks a lot, I'll close this PR and continue on the one in our main repo.

PR #421, opened 2024-11-21
Support nested inline filters

ninarg:
# Why?

We need to support nested inline filters, it's going to be used for "free shipping" filter for recommerce. 

# What?

Currently we support inline filters with 1 level of subfilters only. Nested inline filters are ignored.
Now we want nested filters to be treated like a flat structure. So I manipulated the inline filter by flattening it in init.
Also wrote tests to confirm the expected behaviour. 

A selection of parent and child at the same time should give the same results as selecting child only. 

# Show me

This is how the filter response from backend looks:
```
    {
      "display_name": "Fiks ferdig",
      "name": "shipping_types",
      "filter_items": [
        {
          "display_name": "Fiks ferdig",
          "name": "shipping_types",
          "value": "0",
          "hits": 104,
          "filter_items": [
            {
              "display_name": "Gratis frakt",
              "name": "shipping_types",
              "value": "1",
              "hits": 20,
              "filter_items": [],
              "selected": true
            }
          ],
          "selected": true
        }
      ],
      "type": "STANDARD_FILTER"
    },

```

## Before

The "Gratis frakt" filter is nested, so the filter that should be placed in the same group as "Fiks ferdig" is missing.

## After

https://github.com/user-attachments/assets/5aab38f9-66ad-4003-a7e3-bb89efe92a79

### DIFF
--- a/Sources/Charcoal/CharcoalViewController.swift
+++ b/Sources/Charcoal/CharcoalViewController.swift
@@ -277,7 +277,7 @@ extension CharcoalViewController: FilterViewControllerDelegate {
 
     public func filterViewController(_ viewController: FilterViewController, didSelectFilter filter: Filter) {
         switch filter.kind {
-        case .standard, .freeText, .freeTextOnly:
+        case .standard, .freeText, .freeTextOnly, .inline:
             guard !filter.subfilters.isEmpty else { break }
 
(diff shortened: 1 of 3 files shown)


bstien (approved):
🎉🎉🎉

ninarg:
Closing this, it has been merged into another repo where we'll continue maintaining Charcoal from now on.

